PR #37, opened 2023-04-28
[FOR-1901] Bring the PaymentMethod token calls internal

dleis612:
## What
To keep iOS in line with other SDKs and to prepare for the BT integration, I refactored the balance and capture endpoints to do GET requests to get the PaymentMethod token rather than having the developer store this token and pass it in.

## Test Plan
Added unit tests, adjusted old unit tests, and ran the iOS qa-test suite.


Sample/SampleForageSDK/Sections/RequestBalance/RequestBalanceView.swift
```diff
@@ -150,15 +149,12 @@ class RequestBalanceView: UIView {
...
     // MARK: Private Methods
     
-    private func printPINResult(result: Result<Data?, Error>) {
+    private func printPINResult(result: Result<BalanceModel, Error>) {
         DispatchQueue.main.async {
             switch result {
-            case .success(let data):
-                guard let data = data,
-                      let response = try? JSONDecoder().decode(ForageBalance.self, from: data)
-                else { return }
```
djoksimo, line 159: why did we remove these guards?

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
```diff
@@ -42,9 +42,8 @@ protocol ForageSDKService: AnyObject {
         bearerToken: String,
         merchantAccount: String,
         paymentMethodReference: String,
-        cardNumberToken: String,
         foragePinTextEdit: ForagePINTextField,
-        completion: @escaping (Result<Data?, Error>) -> Void)
+        completion: @escaping (Result<BalanceModel, Error>) -> Void)
```
djoksimo, line 46: I like the use of making these types stricter, like here^ (e.g. `BalanceModel` > `Data`)!

is it safe to no longer have these types marked as optional?

djoksimo: also loving the use of the suffix `Model`^

djoksimo (changes requested):

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
dleis612: It should be fine, the return type was never optional in the first place.

djoksimo (approved):

dleis612:
QA tests and unit tests passed locally for staging.

PR #43, opened 2023-06-13
Add Basistheory support to Pin processing

sgforage:
## What
- Add new abstraction for redundant vault provider
- This PR adds the following abstractions to decouple the vaulting from VGS
  - `VaultResponse` for a common response structure between vgs and basis theory
  - `PINVaultTextField` for representing a common text field that is used by VGS and BT text element wrappers
  - `VaultCollector` to wrap the config and initialization for both BT and VGS
  - `PINVaultTextFieldDelegate` for a common protocol for delgating PIN validation
- Token parsing to ensure that we can parse delimited strings and Xkey alias for both vgs and BT
    
Confirmed that this PR works with the sample app for both VGS and BT. 

## Test Plan


Sources/ForageSDK/Foundation/Network/Collector.swift
```diff
@@ -7,37 +7,166 @@
...
+
+// Wrapper class for VGSCollect
+class VGSCollectWrapper: VaultCollector {
+    
+    public let vgsCollect: VGSCollect
     
-    public static func CreateVGS() -> VGSCollect {
-        return VGSCollect(id: vaultID(ForageSDK.shared.environment).rawValue, environment: environmentVGS(ForageSDK.shared.environment))
+    init(config: VGSCollectConfig) {
+        self.vgsCollect = VGSCollect(id: config.id, environment: config.environment)
```
dleis612, line 37: Important callout that I had to adjust before, we need to always be using a new VGS instance for each input field. Otherwise, two input fields on the same screen will step on each other's toes.

sgforage: i don't think this would conflict with that?

dleis612: Ah yes had this confused with the collector factory function below. You're good.

tioliveira-symphony (commented):
For what I can see:
- It seems fine how it was decoupled. I can't see the details regarding the BasisTheoryElements but seems fine as using the protocol to conform between the wrappers;
- The only thing I would suggest is regarding those force unwrap (eg:. `foragePinTextEdit.collector!`);
- Some indentation throughout the code in general;

Sources/ForageSDK/Foundation/Network/Collector.swift
```diff
@@ -40,4 +146,61 @@ internal class Collector {
...
+    
+    /**
+     BT public Keys
+     */
+    private enum PublicKey: String {
+        case sandbox = "key_DQ5NfUAgiqzwX1pxqcrSzK"
+        case cert = "tntpnht7psv"
+        case prod = "tntbcrncmgi"
+        case staging = "tnteykuh975"
+        case dev = "tntlqkidhc6"
```
djoksimo, line 158: these look like they need to be updated from VGS keys -> BT

Sources/ForageSDK/Foundation/Network/LDManager.swift
```diff
@@ -78,9 +78,9 @@ public class LDManager {
         let vaultPercentage = ld.doubleVariation(forKey: FlagType.vaultPrimaryTrafficPercentage.rawValue, defaultValue: 100.0)
         let randomNum = Double.random(in: 0...100)
         if (randomNum < vaultPercentage) {
-            vaultType = VaultType.btVaultType
-        } else {
             vaultType = VaultType.vgsVaultType
+        } else {
+            vaultType = VaultType.btVaultType
```
djoksimo, line 83: to confirm, VGS is now the primary?^

djoksimo (changes requested):

Sources/ForageSDK/Component/ForagePINTextField/ForagePINTextField.swift
```diff
@@ -96,20 +96,34 @@ public class ForagePINTextField: UIView, Identifiable {
...
+        let vaultType = LDManager.shared.getVaultType()
+        
+        var tf: PINVaultTextField?
+        
+        if (vaultType == VaultType.vgsVaultType) {
+            tf = VGSTextFieldWrapper()
+        } else if (vaultType == VaultType.btVaultType) {
+            tf = BasisTheoryTextFieldWrapper()
+        } else {
+            fatalError("Unexpected vaultType: \(vaultType)")
```
dleis612, line 109: Unlikely it ever happens, but I defaulted to VGS in this scenario. I may add some logging here as well just in case when I implement datadog.

Sources/ForageSDK/Component/ForagePINTextField/Protocol/PINVaultTextField.swift
```diff
@@ -0,0 +1,339 @@
...
+        let regexDigit = try! NSRegularExpression(pattern: "\\d")
+        
+        let pinMask =   [
+            regexDigit,
+            regexDigit,
+            regexDigit,
+            regexDigit
+        ] as [Any]
+        
+        let phoneNumberRegex = try! NSRegularExpression(pattern: "^\\d{4}$")
```
dleis612, line 226: Out of curiosity, why is this called `phoneNumberRegex`?

Sources/ForageSDK/Foundation/Network/Collector.swift
```diff
@@ -40,4 +146,61 @@ internal class Collector {
...
+        case .dev: return .dev
+        }
+    }
+    
+    /**
+     BT proxy Keys
+     */
+    private enum ProxyKey: String {
+        case sandbox = "R1CNiogSdhnHeNq6ZFWrG1"
+        case cert = "cert-proxy-key"
```
dleis612, line 186: Seems like these keys also need to be updated.

Sources/ForageSDK/Foundation/Network/ForageAPI.swift
```diff
@@ -60,7 +60,7 @@ extension ForageAPI: ServiceProtocol {
                 "authorization": "Bearer \(model.authorization)",
                 "content-type": "application/json",
                 "accept": "application/json",
-                "API-VERSION": "2023-03-31"
+                "API-VERSION": "2023-05-15"
```
dleis612, line 63: A rebase on main should pull in the correct `API-VERSION`

dleis612 (changes requested):

Sources/ForageSDK/Component/ForagePINTextField/Protocol/PINVaultTextField.swift
sgforage: cuz i copy-pasta'd too hard from bt docs

Sources/ForageSDK/Foundation/Network/LDManager.swift
sgforage: yes vgs is primary until we start to decide to turn on bt traffic

dleis612 (approved):

djoksimo (approved):

sgforage:
https://github.com/teamforage/forage-ios-sdk/pull/48/files ended up merging here instead due to signing issues

PR #46, opened 2023-06-22
Add token parsing to iOS for BT/VGS

sgforage:
## What
- Add token parsing based on the collector
- Choose the correct X-Key token based on the collector

## Test Plan

- ✅ | ❌ Not a front-end change
- ✅ | ❌ Unit Tests passed locally


Sources/ForageSDK/Foundation/Network/Model/ForageXKeyModel.swift
```diff
@@ -10,4 +10,5 @@ import Foundation
 /// `ForageXKeyModel` used for compose X-key request
 internal struct ForageXKeyModel: Codable {
     let alias: String
+    let bt_alias: String
```
dleis612, line 13: So the encryption_alias endpoint now returns both `alias` and `bt_alias`? Just confirming for the other repos.

dleis612 (approved):

PR #51, opened 2023-07-27
[FX-377] Introduce forageUI protocol and split BT and VGS into separate files

dleis612:
## What
Introduce the ForageUI protocol and prepare for introducing the changes that GoPuff have requested.

## Why
In order to make our solution extensible and prepare for future improvements to the iOS SDK, we will build a protocol that that both the PIN and PAN text fields will inherit from to make their behavior uniform.

## Test Plan
No tests needed as all changes are no-ops.


Sources/ForageSDK/Foundation/Protocol/ForageDelegate.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  File.swift
+//  
+//
+//  Created by Danny Leiser on 7/27/23.
+//
+
+public protocol ForageDelegate: AnyObject {
```
djoksimo, line 8: should be `ForageElementDelegate`

Sources/ForageSDK/Foundation/Protocol/ForageUI.swift
```diff
@@ -0,0 +1,30 @@
...
+    var borderColor: UIColor? { get set }
+    var padding: UIEdgeInsets { get set }
+    var autocorrectionType: UITextAutocorrectionType { get set }
+    var delegate: ForageDelegate? { get set }
+    func isValid() -> Bool
+    func setTranslatesAutoresizingMaskIntoConstraints(_ flag: Bool)
+    func setAccessibilityIdentifier(_ identifier: String)
+    func setIsAccessibilityElement(_ flag: Bool)
+    func setPlaceholderText(_ text: String)
+    func cleanText() -> Void
```
djoksimo, line 28: if we're making breaking changes in this PR, can we rename this to `clearText`

this can be a separate ticket^

Sources/ForageSDK/Foundation/Protocol/ForageUI.swift
```diff
@@ -0,0 +1,30 @@
...
+    func becomeFirstResponder() -> Bool
```
djoksimo, line 29: nice!

Sources/ForageSDK/Foundation/Protocol/ForageUI.swift
```diff
@@ -0,0 +1,30 @@
...
+import UIKit
+
+public protocol ForageUI: UIView {
+    var placeholder: String? { get set }
+    var collector: VaultCollector { get set }
+    var textColor: UIColor? { get set }
+    var tfTintColor: UIColor? { get set }
+    var isSecureTextEntry: Bool { get set }
+    var textAlignment: NSTextAlignment { get set }
+    var font: UIFont? { get set }
```
djoksimo, line 17: do we risk this affecting PIN concealment if I try to change this for the PIN elements?

djoksimo (changes requested):
These changes are awesome! Thank you @dleis612!

devinmorgan:
Hey team, I put [together a proposal](https://www.notion.so/joinforage/ForageUI-Interface-Hierarchy-Proposal-73d686a7bd8c40de8de0e3ab257c946e?pvs=4) for how we may be able to partition many of these attributes into smaller atomic interfaces that could be composed together to form the `ForageUI` (or `ForageElement` if we opt to rename it).

I'll repost in Slack but take a look at how [Stripe organizes their interfaces/config](https://stripe.com/docs/elements/appearance-api?platform=android#:~:text=The%20following%20code%20creates%20the%20appearance%20shown%20in%20the%20screenshot) does for their Android SDK (make sure you click Android tab to see, I think it defaults to Web). I think there is an opportunity to gain some significant developer experience wins by exposing atomic interfaces that can be reused and composed rather than exposing a large monolithic (see the [Interface Segregation Principle](https://www.digitalocean.com/community/conceptual-articles/s-o-l-i-d-the-first-five-principles-of-object-oriented-design#interface-segregation-principle)).

Here is the tl;dr of the interface proposal:
```kotlin
interface ObservableState {
    val isFocused: Boolean
    val isBlured: Boolean
    val isEmpty: Boolean
    val isValid: Boolean
    val isComplete: Boolean
}

interface Appearance {
    var hint: String
    var typeface: Typeface
}

interface Style {
    var textColor: Int
    var textSize: Float
    var hintTextColor: Int
}

interface ForageUI : Appearance, Style, ObservableState {
    fun addFocusEventListener(listener: EventListener)
    fun addBlurEventListener(listener: EventListener)
    fun addChangeEventListener(listener: EventListener)
    fun requestFocus()
}

```

Sources/ForageSDK/Foundation/Protocol/ForageUI.swift
dleis612: Good question I don't believe so, but I think I'm going to remove it for now anyways.

PR #54, opened 2023-07-28
[FX-317] Add ForageLogger

djoksimo:
## What

- Add/install Datadog logging SDK to forage-ios-sdk
- Add interface for handling logs across the SDK 

Note: this does not emit crash reports ; we'll need to enable sentry or [this feature](https://docs.datadoghq.com/real_user_monitoring/error_tracking/ios/?tab=cocoapods)  (under RUM) to send crash reports

TODOs:
- [x] Don't emit logs to datadog when running from simulator
- [x] Don't log to console in the client's debug/console area

## Test Plan

- ✅ Unit Tests passed locally

## How

#55 is based off of this PR 
This is a patch-level change.

Should not be released without #79

### DIFF
--- a/Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
+++ b/Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
@@ -33,8 +33,6 @@ public class DatadogLogger: ForageLogger {
             return
         }
 
-        // TODO: only initialize if not running locally?
-
         Datadog.initialize(
             appContext: .init(),
             trackingConsent: .granted,
@@ -51,7 +49,6 @@ public class DatadogLogger: ForageLogger {
             .sendNetworkInfo(true)
             // how do we decide whether we're in local dev or not?
             .sendLogsToDatadog(true)
-            .printLogsToConsole(true, usingFormat: .shortWith(prefix: "[forage-ios-sdk] "))
             .set(datadogReportingThreshold: .info)
             .build()
     }
--- a/Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
+++ b/Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
@@ -1,0 +1,103 @@
+//
+//  File.swift
+//
+//
+//  Created by Danilo Joksimovic on 2023-07-26.
+//
+
+import Datadog
+import Foundation
+
+// TODO: move this?
+protocol ForageLogger {
+    func setCustomerID(_ customerID: String)
+    func setMerchantAccount(_ merchantAccount: String)
+    func initialize(_ environment: EnvironmentTarget)
+
+    func notice(_ message: String, attributes: [String: Encodable]?)
+    func info(_ message: String, attributes: [String: Encodable]?)
+    func warn(_ message: String, error: Error?, attributes: [String: Encodable]?)
+    func error(_ message: String, error: Error?, attributes: [String: Encodable]?)
+    func critical(_ message: String, error: Error?, attributes: [String: Encodable]?)
+}
+
+public class DatadogLogger: ForageLogger {
+    static let shared = DatadogLogger()
+    private static let DD_CLIENT_TOKEN: String = "pub1e4572ba0f5e53df108c333d5ec66c02"
+    private static var logger: Logger? = nil
+
+    private init() {}
+
+    internal func initialize(_ environment: EnvironmentTarget) {
+        if DatadogLogger.logger != nil {
+            return
+        }
+
+        // TODO: only initialize if not running locally?
+
+        Datadog.initialize(
+            appContext: .init(),
+            trackingConsent: .granted,
+            configuration: Datadog.Configuration
+                .builderUsing(clientToken: DatadogLogger.DD_CLIENT_TOKEN, environment: environment.rawValue)
+                .set(serviceName: "ios-sdk")
+                .set(endpoint: .us1)
+                .build()
+        )
+
+        Datadog.verbosityLevel = .info
+
+        DatadogLogger.logger = Logger.builder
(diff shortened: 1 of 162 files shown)


djoksimo:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/54" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/54?utm_source=stack-comment).

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -0,0 +1,102 @@
...
+    private static var logger: Logger? = nil
+
+    private init() {}
+
+    internal func initialize(_ environment: EnvironmentTarget) {
+        if DatadogLogger.logger != nil {
+            return
+        }
+
+        // TODO: only initialize if not running locally?
```
dleis612, line 36: I think we want to emit logs while running locally. The only scenario I think we should avoid logging is when we are running unit tests.

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -0,0 +1,102 @@
...
+                .build()
+        )
+
+        Datadog.verbosityLevel = .info
+
+        DatadogLogger.logger = Logger.builder
+            .sendNetworkInfo(true)
+            // how do we decide whether we're in local dev or not?
+            .sendLogsToDatadog(true)
+            .printLogsToConsole(true, usingFormat: .shortWith(prefix: "[forage-ios-sdk] "))
```
dleis612, line 54: I decided against enabling this field. I don't want to clog up client logs with a ton of our statements, we just want to be able to emit them to our own datadog instance.

dleis612 (changes requested):

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -0,0 +1,102 @@
...
+            appContext: .init(),
+            trackingConsent: .granted,
+            configuration: Datadog.Configuration
+                .builderUsing(clientToken: DatadogLogger.DD_CLIENT_TOKEN, environment: environment.rawValue)
+                .set(serviceName: "ios-sdk")
+                .set(endpoint: .us1)
+                .build()
+        )
+
+        Datadog.verbosityLevel = .info
```
devinmorgan, line 48: Curious why we want to do the initialization of the Datadog singleton inside the instance method of `initialize` instead of somewhere more upstream in the application? From what I can tell, this initialization-related code is not something that `DatadogLogger` cares about. Instead, `DatadogLogger` seems only to care that `Datadog` was indeed initialized **before** an instance of `DatadogLogger` is created. 

What do you think about a pattern like:
```swift
// somewhere in the app initialization
Datadog.initialize(
    appContext: .init(),
    trackingConsent: .granted,
    configuration: Datadog.Configuration
        .builderUsing(clientToken: DatadogLogger.DD_CLIENT_TOKEN, environment: environment.rawValue)
        .set(serviceName: "ios-sdk")
        .set(endpoint: .us1)
        .build()
)
                
// in ForageLogger.swift
public class DatadogLogger: ForageLogger {
    // not shared since its not static
    private var logger: Logger? = nil

    private init(...loggerRelatedConfig) {
        this.logger = Logger.builder
            // apply logerRelatedConfig
            .build()
    }

    // no more initialize function since all of the async setup is done before
}
```

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    internal func setPaymentRef(_ paymentRef: String) {
+        setLogAttribute("payment_ref", value: paymentRef)
+    }
+    
+    internal func setPaymentMethodRef(_ paymentMethodRef: String) {
+        setLogAttribute("payment_method_ref", value: paymentMethodRef)
+    }
+
```
devinmorgan, line 78: Do we anticipate needing to update these values on the same `ForageLogger` instance? Is there an opportunity to reduce complexity by passing these as a `Config` object to `init(config: Config)`?

devinmorgan (commented):
Danilo - I appreciate the work here. A couple of things I wanted to mention:
1. We were discussing some [interface designs for the Mobile SDKs](https://www.notion.so/joinforage/Mobile-SDK-Logger-Interface-Proposal-f97d859471a0424c83d19e46665aa1f9?pvs=4). You are more familiar with the intricacies of iOS and Forage's iOS than I am, so if the interfaces we've discussed don't make sense to implement right now, that is OK.
2. I had a few design-related questions I wanted to clarify before approving. Let me know what you think or if you want to hop a call to discuss happy to do that as well

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
djoksimo: 657bf39a9c0b05d39ce7e681271fb19e7bd8447c

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
djoksimo: 657bf39a9c0b05d39ce7e681271fb19e7bd8447c

devinmorgan (approved):
Thanks for considering my comments! This latest implementation strikes a balance of speed while laying the ground for proper Logger service via the `ForagerLogger` protocol. Approved!

dleis612 (approved):

PR #57, opened 2023-08-02
[FX-269] Add support for optional reusable param

djoksimo:
## What

- Add optional reusable param to README
- Add/display reusable value to Card Number View

### TODO
- [ ] update unit tests - will add them after #63 gets merged.

## Why

@devinmorgan  I considered moving [individual params (reusable, customerID, merchantAccount) to an object](https://linear.app/joinforage/issue/FX-363/android-move-reusable-and-customerid-args-into-optionsconfig) – but the position of the params will still matter even if we separate the params into a separate struct

image

## Test Plan

- Test when reusable is missing from the call to `tokenizeEBTCard`, `reusable` is true ✅ 
- Test when reusable is true, is `reusable` true in the response (in the Card Number View) ✅ 

- Test when reusable is false, is `reusable` false... ✅ 

- ✅ Unit Tests passed locally

## How

Can be released as-is. This is a minor-level version change since the reusable flag is optional

### DIFF
--- a/Sample/SampleForageSDK/Sections/CardNumber/CardNumberView.swift
+++ b/Sample/SampleForageSDK/Sections/CardNumber/CardNumberView.swift
@@ -142,6 +142,17 @@ class CardNumberView: UIView {
         return label
     }()
     
+    private let reusableLabel: UILabel = {
+        let label = UILabel()
+        label.text = ""
+        label.textColor = .black
+        label.font = UIFont.systemFont(ofSize: 14, weight: .bold)
+        label.numberOfLines = 0
+        label.accessibilityIdentifier = "lbl_reusable"
+        label.isAccessibilityElement = true
+        return label
+    }()
+    
     private let errorLabel: UILabel = {
         let label = UILabel()
         label.text = ""
@@ -159,7 +170,8 @@ class CardNumberView: UIView {
         ForageSDK.shared.tokenizeEBTCard(
             bearerToken: ClientSharedData.shared.bearerToken,
             merchantAccount: ClientSharedData.shared.merchantID,
-            customerID: ClientSharedData.shared.customerID) { result in
+            customerID: ClientSharedData.shared.customerID,
+            reusable: ClientSharedData.shared.reusable) { result in
                 self.printResult(result: result)
             }
     }
@@ -187,6 +199,8 @@ class CardNumberView: UIView {
                 self.tokenLabel.text = "token=\(response.card.token)"
                 self.last4Label.text = "last4=\(response.card.last4)"
                 self.customerIDLabel.text = "customerID=\(response.customerID ?? "NO CUST ID")"
+                self.reusableLabel.text = "reusable=\(response.reusable)"
+
                 self.errorLabel.text = ""
                 ClientSharedData.shared.paymentMethodReference = response.paymentMethodIdentifier
                 self.updateButtonState(isEnabled: true, button: self.nextButton)
@@ -197,6 +211,7 @@ class CardNumberView: UIView {
                 self.tokenLabel.text = ""
                 self.last4Label.text = ""
                 self.customerIDLabel.text = ""
+                self.reusableLabel.text = ""
                 self.updateButtonState(isEnabled: false, button: self.nextButton)
             }
             
@@ -215,6 +230,7 @@ class CardNumberView: UIView {
         contentView.addSubview(tokenLabel)
         contentView.addSubview(last4Label)
         contentView.addSubview(customerIDLabel)
+        contentView.addSubview(reusableLabel)
         contentView.addSubview(errorLabel)
         contentView.addSubview(sendPanButton)
         contentView.addSubview(nextButton)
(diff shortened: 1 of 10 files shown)


djoksimo:
* main
  * **PR #57** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/57" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/57?utm_source=stack-comment).

devinmorgan:
> ## Why
> @devinmorgan I considered moving [individual params (reusable, customerID, merchantAccount) to an object](https://linear.app/joinforage/issue/FX-363/android-move-reusable-and-customerid-args-into-optionsconfig) – but the position of the params will still matter even if we separate the params into a separate struct
> 
> image


Danilo! Thanks for calling out about the `ForageConfig`. After [poking around on Stack Overflow](https://stackoverflow.com/a/37553704/10541855), are we saying that something like this would not give us the desired optional-param-where-order-doesn't-matter behavior?
```
struct ForageConfig {
   var reusable : Bool
   var customerID : String?

   init(reusable: Bool? = true, customerID: String? = nil) {
       self.reusable = reusable!
       self.customerID = customerID
   }
}

let configWithCustomerId = ForageConfig(customerID: "12345")
let configWithReusable = ForageConfig(reusable: false)
```

djoksimo:
> > ## Why
> > @devinmorgan I considered moving [individual params (reusable, customerID, merchantAccount) to an object](https://linear.app/joinforage/issue/FX-363/android-move-reusable-and-customerid-args-into-optionsconfig) – but the position of the params will still matter even if we separate the params into a separate struct
> > image
> 
> Danilo! Thanks for calling out about the `ForageConfig`. After [poking around on Stack Overflow](https://stackoverflow.com/a/37553704/10541855), are we saying that something like this would not give us the desired optional-param-where-order-doesn't-matter behavior? I don't have Xcode on my machine, so regretfully am nothing more than a bystander with questions, lol.
> 
> ```swift
> struct ForageConfig {
>    var reusable : Bool
>    var customerID : String?
> 
>    init(reusable: Bool? = true, customerID: String? = nil) {
>        self.reusable = reusable!
>        self.customerID = customerID
>    }
> }
> 
> let configWithCustomerId = ForageConfig(customerID: "12345")
> let configWithReusable = ForageConfig(reusable: false)
> ```

The order of the params will still matter in this case and the compiler will yell with an error if you violate the struct initialization ordering.

dleis612 (approved):

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
devinmorgan, line 1: I do want to flag that a function requiring more than 4 arguments is generally considered [bad practice in Swift](https://rules.sonarsource.com/swift/RSPEC-107/). I know that we had been playing around with a `ForageConfig` object and that while we were able to make values optional, the compiler enforced the parameters be at least passed in the order they were provided (here's [a Repl.it demo](https://replit.com/@devinmorgan3/Swift#main.swift)). Enforcing order sounds OK to me but I could be missing something.

I think the main point here is that it's likely that we're going to want the ability to continue adding more parameters to these submit-methods in the future so we should come up with an answer to keep anti-patterns at bay. 

Fwiw, another option seems to be the builder pattern which seems to be very common in OO languages like Swift and Kotlin

```swift
class ForageConfigBuilder {
    private var reusable: Bool = true
    private var customerID: String?

    func setReusable(_ reusable: Bool) -> ForageConfigBuilder {
        self.reusable = reusable
        return self
    }

    func setCustomerID(_ customerID: String) -> ForageConfigBuilder {
        self.customerID = customerID
        return self
    }

    func build() -> ForageConfig {
        return ForageConfig(reusable: reusable, customerID: customerID)
    }
}

struct ForageConfig {
    let reusable: Bool
    let customerID: String?

    init(reusable: Bool, customerID: String?) {
        self.reusable = reusable
        self.customerID = customerID
    }
}

// using it like this
let config = ForageConfigBuilder()
    .setReusable(false)
    .setCustomerID("12345")
    .build()
    
tokenizeEBTCard(
    bearerToken,
    merchantAccount,
    customerID,
    config
```

though if we're going to use the builder pattern at all, we probably want to put all function arguments inside of it and have `tokenizeEBTCard(config)`

djoksimo: @devinmorgan thank you for bringing this up.

We should ideally move the `bearerToken` and `merchantAccount` args into the `ForageSDK.setup()` instead of having to pass them in each submit method. 

I believe this can be a clean up ticket and doesn't have to block the release of this PR (as `tokenizeEBTCard` takes 4 params, one of which is optional).

We have room to make some additional breaking changes. I've created a ticket for this and will add it to the next cycle

devinmorgan: Your comment here, plus the conversation we had on standup where it surfaced that only a few other folks besides GoPuff will be using the Android SDK any time soon, + our good relationship with them, suggests we'll have the opportunity to restructure this code later. So, this PR should not be blocked because we want to get the raw functionality out there. Thanks for the fruitful discussion, Danilo!

Sample/SampleForageSDK/Sections/CardNumber/CardNumberView.swift
```diff
@@ -319,6 +336,15 @@ class CardNumberView: UIView {
...
         
+        errorLabel.anchor(
+            top: reusableLabel.safeAreaLayoutGuide.bottomAnchor,
+            leading: contentView.safeAreaLayoutGuide.leadingAnchor,
+            bottom: nil,
+            trailing: contentView.safeAreaLayoutGuide.trailingAnchor,
+            centerXAnchor: contentView.centerXAnchor,
+            padding: UIEdgeInsets(top: 24, left: 24, bottom: 0, right: 24)
+        )
+
```
devinmorgan, line 347: Were these updates to support QA tests and/or to help with local development? Good idea! I should take a hint from this in Android

devinmorgan (approved):
Thanks for the fruitful discussion before - looks great!

Sample/SampleForageSDK/Sections/CardNumber/CardNumberView.swift
djoksimo: yep, both local dev and qa tests^

dleis612 (approved):

PR #59, opened 2023-08-04
[FX-387] Implement onFocusChange and onTextChange for PAN

dleis612:
## What
Have the PAN element implement `ForageElement` and wire up the `onFocusChange` and `onTextChange` events.

## Test Plan
Added unit tests to check new implementation and adjust qa-tests to check for the updated state values that are displayed in the sample app.

https://github.com/teamforage/forage-ios-sdk/assets/89605898/3808cc29-40d2-45c3-8199-cfa3ef3d5d48

### DIFF
--- a/Sources/ForageSDK/Component/ForagePANTextField/ForagePANTextField.swift
+++ b/Sources/ForageSDK/Component/ForagePANTextField/ForagePANTextField.swift
@@ -8,21 +8,61 @@
 import UIKit
 import VGSCollectSDK
 
-private enum CardType: String {
-    case ebt = "ebt"
+public enum CardType: String {
+    case EBT = "ebt"
 }
 
-public class ForagePANTextField: UIView, Identifiable {
+public class ForagePANTextField: UIView, Identifiable, ForageElement {
+    public func setPlaceholderText(_ text: String) {
+        
+    }
+    
+    public func clearText() {
+        ForageSDK.shared.panNumber = ""
+    }
+    
+    /// BorderWidth for the text field
+    private var _borderWidth: CGFloat = 0.1
+    @IBInspectable public var borderWidth: CGFloat {
+        get { return _borderWidth }
+        set { _borderWidth = newValue }
+    }
+    
+    /// BorderColor for the text field
+    private var _borderColor: UIColor? = .black
+    @IBInspectable public var borderColor: UIColor? {
+        get { return _borderColor }
+        set { _borderColor = newValue }
+    }
+    
+    /// Padding for the text field
+    private var _padding: UIEdgeInsets = UIEdgeInsets(top: 0.0, left: 0.0, bottom: 0.0, right: 0.0)
+    @IBInspectable public var padding: UIEdgeInsets {
+        get { return _padding }
+        set { _padding = newValue }
+    }
+    
+    @IBInspectable public var isEmpty: Bool {
+        get { return ForageSDK.shared.panNumber.isEmpty }
+    }
+    
+    private var _isValid: Bool = true
+    @IBInspectable public var isValid: Bool {
+        get { return _isValid }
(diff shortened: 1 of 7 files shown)


devinmorgan (approved):
This man ships! 🚢

Sources/ForageSDK/Component/ForagePANTextField/ForagePANTextField.swift
```diff
@@ -218,54 +252,72 @@ public class ForagePANTextField: UIView, Identifiable {
 // MARK: - UITextFieldDelegate
 
 extension ForagePANTextField: UITextFieldDelegate {
+    public func textFieldDidBeginEditing(_ textField: UITextField) {
+        delegate?.focusDidChange(self)
+    }
+    
+    public func textFieldDidEndEditing(_ textField: UITextField) {
```
devinmorgan, line 259: My understanding is that the `textFieldDidEndEditing` handler is equivalent to [a blur event](https://developer.apple.com/documentation/uikit/uitextfielddelegate/1619591-textfielddidendediting). @dleis612 was this intentional? I want to make sure that this can't cause GoPuff to misread a blur event as a focus event and cause a wonky experience for users.

Sources/ForageSDK/Component/ForagePANTextField/ForagePANTextField.swift
```diff
@@ -218,54 +252,72 @@ public class ForagePANTextField: UIView, Identifiable {
...
+        delegate?.focusDidChange(self)
+    }
```
devinmorgan, line 261: My understanding is that the `textFieldDidEndEditing` handler is equivalent to [a blur event](https://developer.apple.com/documentation/uikit/uitextfielddelegate/1619591-textfielddidendediting). @dleis612 was this intentional? I want to make sure that this can't cause GoPuff to misread a blur event as a focus event and cause a wonky experience for users.

dleis612: It is equivalent yes. The event handler is just a general `onFocusDidChange` which encompasses losing focus and gaining focus right?

devinmorgan: I agree that the name `onFocusDidChange` expresses that it could be called for `blur` and `focus` events. That said, I would think we would want to give SDK users the ability to distinguish between `blur` and `focus` events.

The specific scenario I am concerned about is if GoPuff is doing their own logic to trigger the appearance of the keyboard on focus, then their logic would fire inappropriately if they use our `onFocusDidChange` event subscriber because they'll attempt to launch the keyboard again when the user is trying to leave the input field.

Said differently, I would have thought something like these would offer the information SDK developers would care to know:

**A single method that invokes the same callback but includes a boo to distinguish `focus` from `blur`**
```swift
public func textFieldDidBeginEditing(_ textField: UITextField) {
    delegate?.focusDidChange(self, hasFocus: true) // focus event
}

public func textFieldDidEndEditing(_ textField: UITextField) {
    delegate?.focusDidChange(self, hasFocus: false) // blur event
}
```

**Two distinct methods that invoke either `focus` xor `blur` callback**
```swift
 public func textFieldDidBeginEditing(_ textField: UITextField) {
    delegate?.onFocusEvent(self) // focus event
}

public func textFieldDidEndEditing(_ textField: UITextField) {
    delegate?.onBlurEvent(self) // blur event
}
```

What is your take?

dleis612: For now, the implemented logic should work fine. The event handler returns the state object that has either `isFirstResponder=true` or `isFirstResponder=false`, which they can check to determine if they have focus or not.

PR #60, opened 2023-08-07
[FX-392] Move merchantAccount + bearerToken to Config

djoksimo:
## What

- Move `merchantAccount` and `bearerToken` from the submit methods to the `ForageSDK.Config`
- Call `ForageSDK.setup()` with the new `ForageSDK.Config` to update the environment, session token, merchant account, and/or bearer token

## Why

https://linear.app/joinforage/issue/FX-392/move-merchantaccount-bearertoken-to-foragesdksetup-in-forage-ios-sdk

Opted against using `@available` and gracefully deprecating the old method signatures (make the change reverse compatible), since only one client will have to make this small + trivial change on their end and we're currently making other breaking changes to the SDK anyway.

## Test Plan

- ✅ Unit Tests passed locally

## How

- Breaking change!
- [ ] Merge down this PR #75 
- [ ] Merge down PR #86
- [ ] Merge down PR #86 
- [ ] Merge down PR #87 
- [ ] Merge down PR #88 
- [ ] Communicate this change and the [migration guide](https://www.notion.so/joinforage/iOS-SDK-Migration-guide-v3-v4-86907ee4b4e34caba9633dc3774cca97) to the client when we release this version
- [ ] Update the [readme docs](https://docs.joinforage.app/docs/ios#starter-code)

### Migration guide (v3 -> v4)

Can be [found in Notion](https://www.notion.so/joinforage/iOS-SDK-Migration-guide-v3-v4-86907ee4b4e34caba9633dc3774cca97)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,21 +2,27 @@
 
 The ForageSDK processes Electronic Benefits Transfer (EBT) payments in your e-commerce application. It provides secure user interfaces for collecting and tokenizing an EBT cardholder's PAN and accepting an EBT cardholder's PIN to execute a balance check and process a payment.
 
-Table of contents
-=================
+# Table of contents
 
 <!--ts-->
-   * [Integration](#integration)
-      * [CocoaPods](#cocoapods)
-      * [Swift Package Manager](#swift-package-manager) 
-   * [Usage](#usage)
-      * [Create ForageSDK instance](#create-foragesdk-instance)
-      * [Forage UI Elements](#forage-ui-elements)
-      * [ForagePANTextField](#foragepantextfield)
-      * [ForagePINTextField](#foragepintextfield)
-      * [Demo Application](#demo-application)
-   * [Dependencies](#dependencies)
-<!--te-->
+
+- [ForageSDK](#foragesdk)
+- [Table of contents](#table-of-contents)
+  - [Integration](#integration)
+    - [CocoaPods](#cocoapods)
+    - [Swift Package Manager](#swift-package-manager)
+    - [Step-by-step](#step-by-step)
+  - [Usage](#usage)
+    - [Import SDK into your file](#import-sdk-into-your-file)
+    - [Create ForageSDK instance](#create-foragesdk-instance)
+  - [Forage UI Elements](#forage-ui-elements)
+    - [ForagePANTextField](#foragepantextfield)
+    - [Tokenize card number](#tokenize-card-number)
+    - [ForagePINTextField](#foragepintextfield)
+    - [Balance](#balance)
+    - [Capture payment](#capture-payment)
+  - [Demo Application](#demo-application)
+  - [Dependencies](#dependencies)
 
 ## Integration
 
@@ -58,6 +64,7 @@ Click on `Add Package` button. And, on the next screen, select the package `Fora
 ## Usage
 
 ### Import SDK into your file
+
 ```swift
 import ForageSDK
 ```
@@ -68,7 +75,11 @@ To initialize a ForageSDK instance, you need to provide the environment.
 
 ```swift
 ForageSDK.setup(
-    ForageSDK.Config(environment: .sandbox)
+    ForageSDK.Config(
+        environment: .sandbox,
+        merchantAccount: "mid/abcd123",
+        bearerToken: "sandbox_eyJ0eXAiOiJKV1Qi..."
+    )
 )
 ```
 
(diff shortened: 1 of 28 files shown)


djoksimo:
* main
  * **PR #60** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/60" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/60?utm_source=stack-comment).

devinmorgan:
> Opted against using @available and gracefully deprecating the old method signatures (make the change reverse compatible), since only one client will have to make this small + trivial change on their end and we're currently making other breaking changes to the SDK anyway.

This is the right call 👍

devinmorgan (approved):
Nice work Danilo - thanks for updating the README accordingly!

dleis612 (commented):
Not sure this is something we want to do. On the one side, this would be great for GoPuff because they won't ever need to change either value. On the other side, there will likely be future Merchants who want to pass in a different `merchantAccount` value based on which store you shop for.

We can tell them to pass in a new Forage Config, but due to the static nature of things in iOS that wouldn't work either since we set the config once and never run that code again.

djoksimo:
> Not sure this is something we want to do. On the one side, this would be great for GoPuff because they won't ever need to change either value. On the other side, there will likely be future Merchants who want to pass in a different `merchantAccount` value based on which store you shop for.
> 
> We can tell them to pass in a new Forage Config, but due to the static nature of things in iOS that wouldn't work either since we set the config once and never run that code again.

Discussed via DM and Forage X office hours; The client can call `ForageSDK.setup` multiple times; I'll add some tests for this as well.

```tsx
ForageSDK.setup(Config(merchantID: firstMerchantID))
ForageSDK.setup(Config(merchantID: newMerchantID)) // update merchantID
```

README.md
```diff
@@ -68,7 +75,11 @@ To initialize a ForageSDK instance, you need to provide the environment.
 
 ```swift
 ForageSDK.setup(
-    ForageSDK.Config(environment: .sandbox)
+    ForageSDK.Config(
+        environment: .sandbox,
+        merchantID: "mid/abcd123",
```
dleis612, line 80: Let's use the old pattern here still. `merchantID: "1234567"`

Sample/SampleForageSDK/Sections/CreatePayment/CreatePaymentView.swift
```diff
@@ -235,7 +235,7 @@ class CreatePaymentView: UIView {
                 self.fundingTypeLabel.text = "fundingType=\(response.fundingType.rawValue)"
                 self.paymentMethodIdentifierLabel.text = "paymentMethodIdentifier=\(response.paymentMethodIdentifier)"
                 self.paymentIdentifierLabel.text = "paymentIdentifier=\(response.paymentIdentifier)"
-                self.merchantAccountLabel.text = "merchantAccount=\(response.merchantAccount)"
+                self.merchantIDLabel.text = "merchantID=\(response.merchantID)"
```
dleis612, line 238: I have a feeling this is selector is going to need to be adjusted in the qa-tests.

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -40,49 +42,65 @@ public class ForageSDK {
...
+             merchantID: "mid/abcd123",
+             sessionToken: "sandbox_eyJ0eXAiOiJKV1Qi..."
+         )
+     )
     ````
      */
     public class func setup(_ config: Config) {
         ForageSDK.config = config
+        ForageSDK.shared.environment = config.environment
+        ForageSDK.shared.merchantAccount = config.merchantAccount
```
dleis612, line 103: Is this supposed to be `merchantId` and the line below `sessionToken`?

dleis612 (dismissed):

README.md
djoksimo: New SDK users are instructed to use the "Merchant ID" across our docs, which has this format. Are we not fully set on this format or am I missing some other context?

Sample/SampleForageSDK/Sections/CreatePayment/CreatePaymentView.swift
djoksimo: QA tests passed; I don't think that they rely on the naming for this^

README.md
dleis612: That was changed to return to the original format. We aren't pushing that format until we have food as medicine customers.

devinmorgan (approved):
Looked this PR over again, and it makes sense so giving it a ✅

PR #64, opened 2023-08-07
[POP-69] Enforce 80% code coverage for PRs

nestorga:
## What
This replicates the behavior of [forage-android-sdk/#51](https://github.com/teamforage/forage-android-sdk/pull/51) but for our iOS SDK. 

Coverage is enforced at 80% for lines changed in the PR via a three-step process:

- First, the raw code coverage data is generated from `xcodebuild`.
- Next, XCode's proprietary format is converted to XML using `slather`.
- Finally, we use `diff-cover` just as in Android to enforce code coverage. If the changed lines have insufficient coverage, it will fail the status check. 

Just as with Android, this won't actually block PRs unless we set it up as a `main` branch required status check.

Also of note, this PR adds the `ForageSDK` scheme as an explicit scheme so that it can be picked up by `xcodebuild` at the CLI in GH. Previously, the scheme was auto-created by XCode and I simply added it explicitly to the project (Under "Manage Schemes" > Set "Container" to "SampleForageSDK project"). Then this scheme auto-creates its test plan which is what actually runs the test suite. This also necessitated removing `.xcodeproj` from `.gitignore`.

## Why
We want to begin improving code coverage in the iOS SDK.

## Test Plan
- Verified that the workflow works as expected on GitHub
- Verified locally that missed lines in a PR are correctly reported by `diff-cover`. A previous approach had a bug in the XCode->XML translation step that broke this. 

## How
We can merge this without affecting ongoing development activity as noted above.


devinmorgan (approved):
Couldn't tell if this was still WIP or not so figure I give you approval just in case!

Gemfile
```diff
@@ -0,0 +1,5 @@
+source "https://rubygems.org"
+
+gem "xcov"
```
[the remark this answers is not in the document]
nestorga: This is not implemented yet. Will ping you when it's ready for review

nestorga:
Good morning @devinmorgan,

Typically I use the PR draft state + comments to indicate if my PR is WIP or not:

Screenshot 2023-08-08 at 9 03 20 AM

Once ready for review, I'll move it from "Draft" to "Open" and ping you in the CR channel. Normally I'd wait to add reviewers, but it's enforced by `CODEOWNERS` in this repo.

devinmorgan:
> Normally I'd wait to add reviewers, but it's enforced by CODEOWNERS in this repo.

Ahhh OK yeah sounds like this is why I got pinged for the review. Cool beans - looking forward to what you ship next!

.github/workflows/Coverage.yaml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/setup-python@v4 
+        with:
+          python-version: '3.9'
+          cache: 'pip'
+
+      - name: Install Python dependencies
+        run: pip install -r requirements.txt
+
+      - name: Install Slather
+        run: gem install slather
```
nestorga, line 27: Previously, I tried using a `Gemfile` and the [setup-ruby](https://github.com/ruby/setup-ruby) action. Unfortunately, this didn't modify the `PATH` as promised so I fell back to this comparatively simple approach.

nestorga, line 27: Previously, I tried using a `Gemfile` and the [setup-ruby](https://github.com/ruby/setup-ruby) action. Unfortunately, this didn't modify the `PATH` as promised so I fell back to this comparatively simple approach. 

I also tried `gem install slather -v 1.8.1`, but this seems to have caused all the dependencies to be built which adds several minutes to the build time, versus this which takes 5 seconds.

Given the difficulties above, I think using latest version here is the best option.

.github/workflows/Coverage.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Run tests and generate coverage report (XCode)
+        run: |
+          xcodebuild test \
+            -project Sample/SampleForageSDK.xcodeproj \
+            -scheme ForageSDK \
+            -destination 'platform=iOS Simulator,name=iPhone 14 Pro,OS=latest' \
+            -enableCodeCoverage YES
```
nestorga, line 35: There was a [mysterious run ](https://github.com/teamforage/forage-ios-sdk/actions/runs/5802312850/job/15728454902)where this timed out after 10 minutes. Usually it takes around 5. Unsure if this is related to GH or the build itself. 

Something to keep an eye on.

.github/workflows/Coverage.yaml
```diff
@@ -0,0 +1,48 @@
+name: Coverage
+
+on: [pull_request]
+
+jobs:
+  coverage:
+    runs-on: macos-latest
```
owenkim, line 7: This increases the cost by 10x compared to Linux. Are we sure we want to do this? Maybe just for every merge instead of every PR update?

owenkim (changes requested):
I think it's worth thinking through the cost here before we hop into it.

.github/workflows/Coverage.yaml
owenkim: Nevermind. https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions 

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners

owenkim (approved):

PR #55, opened 2023-07-28
[FX-317] Part 2: Add log statements

djoksimo:
## What

- Add log statements with helpful attributes (context), prefixes
- Logs all HTTP requests and responses
- Logs all errors with level "error"

## Why

Learn more: https://docs.google.com/document/d/1NYe6JtnBJAVrtjIKREdwoBt3lqh18vfXqzj9EaodXhI/edit#bookmark=id.2rdd25jqml4p

## Test Plan

DD logs https://app.datadoghq.com/logs?query=service%3Aios-sdk%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1690904666204&to_ts=1690904713184&live=false

image

- ✅  iOS QA Tests passed locally
- ✅  Unit Tests passed locally

## How

This PR Stack (FX-317) can be considered a patch-level change.
This PR is based off of #54

PR #65, opened 2023-08-08
[FX-394] Add vault metrics logs 

djoksimo:
## What

View dashboards [here in Datadog](https://app.datadoghq.com/dashboard/82m-czd-39c/vaulting-metrics?tpl_var_env%5B0%5D=sandbox&from_ts=1693243688272&to_ts=1693244588272&live=true)

View [related metrics logs ](https://app.datadoghq.com/logs?query=service%3Aios-sdk%20log_kind%3Ametric%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1693244194015&to_ts=1693244599810&live=false) (and their context/attributes)

TODOs:
- [x] Dashboards for error count (last 15 minutes) and error rate (errors / attempts)
- [x] Complete rest of dashboards: https://app.datadoghq.com/dashboard/82m-czd-39c/vaulting-metrics?from_ts=1692199494998&to_ts=1692200394998&live=true

image

## Why

https://linear.app/joinforage/issue/FX-394/ios-add-latency-vault-type-response-code-to-dd-logs

## Test Plan

- Added unit tests
- Observed DD logs, metrics explorer and [dashboard graphs](https://app.datadoghq.com/dashboard/82m-czd-39c/vaulting-metrics?tpl_var_env%5B0%5D=sandbox&from_ts=1693243688272&to_ts=1693244588272&live=true) for BT and VGS

- ✅ Unit Tests passed locally

## How

Can be released as-is


djoksimo:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/54" target="_blank">Graphite</a> 
      * **PR #65** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/65" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/65?utm_source=stack-comment).

djoksimo:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/54" target="_blank">Graphite</a> 
      * **PR #65** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/65" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/65?utm_source=stack-comment).

Sources/ForageSDK/Foundation/Network/Collector.swift
```diff
@@ -12,10 +12,11 @@ import BasisTheoryElements
 let tokenDelimiter = ","
 let tokenKey = "card_number_token"
 
-public protocol VaultCollector {
+internal protocol VaultCollector {
```
djoksimo, line 15: not sure why this was public before?

@dleis612

dleis612: uhhhh definitely want this to be internal.

dleis612 (approved):

Sources/ForageSDK/Foundation/Telemetry/Metrics.swift
```diff
@@ -0,0 +1,39 @@
...
+            prefix: "Metrics"
+        )
+    )
+    let vaultName = vaultType.rawValue
+    metricsLogger?.setLogKind(ForageLogKind.metric).info("Received response from \(vaultName) proxy", attributes: [
+        "vault_type": vaultName,
+        "path": latencyAttributes.path,
+        "method": latencyAttributes.method.rawValue,
+        "http_status": latencyAttributes.httpStatusCode,
+        "roundtrip_ms": latencyAttributes.roundTripMs,
```
sgforage, line 36: can we call it latency to standardize on names?

sgforage (approved):

Sources/ForageSDK/Foundation/Telemetry/Metrics.swift
devinmorgan, line 1: Hey Danilo! What do you think of making this `struct` and these methods a class instead? (see below)

The current design requires making all consumers of this code aware of their implementation details like `DispatchTime` and even calculations/function calls (e.g. `calculateTimeDifferenceInMilliseconds`), whereas I think we may be able to hide a lot of these details away to make it easier to maintain/evolve.

```swift
// Base Protocol
protocol Tracer {
    func start()
    func end()
    func captureAsLog()
}

// Extended Protocol for network related tracing
protocol NetworkTracer: Tracer {
    func setPath(path: String)
    func setMethod(method: HttpMethod)
    func setHttpStatusCode(code: Int)
}

class VaultProxyResponseTracer: NetworkTracer {
    
    private var startTime: DispatchTime?
    private var endTime: DispatchTime?
    
    private var path: String?
    private var method: HttpMethod?
    private var httpStatusCode: Int?
    private let action: VaultAction
    private let metricsLogger: ForageLogger? = DatadogLogger(ForageLoggerConfig(prefix: "Metrics"))
    
    private init(action: VaultAction) {
        self.action = action
    }
    
    static func newTraceForVaultAction(action: VaultAction) -> VaultProxyResponseTracer {
        return VaultProxyResponseTracer(action: action)
    }
    
    func start() {
        startTime = DispatchTime.now()
    }
    
    func end() {
        endTime = DispatchTime.now()
    }
    
    func setPath(path: String) {
        self.path = path
    }
    
    func setMethod(method: HttpMethod) {
        self.method = method
    }
    
    func setHttpStatusCode(code: Int) {
        self.httpStatusCode = code
    }
    
    func captureAsLog() {
        guard let startTime = startTime, let endTime = endTime, let path = path, let method = method else {
            print("Incomplete tracer data. Can't capture log.")
            return
        }
        
        let roundTripMs = calculateTimeDifferenceInMilliseconds(from: startTime, to: endTime)
        let vaultName = action.rawValue
        
        metricsLogger?.setLogKind(ForageLogKind.metric).info("Received response from \(vaultName) proxy", attributes: [
            "vault_type": vaultName,
            "path": path,
            "method": method.rawValue,
            "http_status": httpStatusCode ?? "N/A",
            "roundtrip_ms": roundTripMs,
            "action": action.rawValue
        ])
    }
    
    private func calculateTimeDifferenceInMilliseconds(from startTime: DispatchTime, to endTime: DispatchTime) -> Double {
        let nanoseconds = endTime.uptimeNanoseconds - startTime.uptimeNanoseconds
        let milliseconds = Double(nanoseconds) / 1_000_000
        return milliseconds
    }
}
```

Sources/ForageSDK/Foundation/Network/Collector.swift
```diff
@@ -49,7 +55,7 @@ class VGSCollectWrapper: VaultCollector {
         vgsCollect.customHeaders = mutableHeaders
     }
     
-    func sendData(path: String, extraData: [String: Any], completion: @escaping (VaultResponse) -> Void) {
+    func sendData(path: String, vaultAction: VaultAction, extraData: [String: Any], completion: @escaping (VaultResponse) -> Void) {
```
devinmorgan, line 58: > 💡  NOTE: This comment references a comment that I made on `Metrics.swift`.

It seems like there may be an opportunity to hide a lot of the implementation details about how we collect start/end timestamps and even the details that we collect in the trace/log by passing a class instance of `VaultProxyResponseTracer` to `sendData` instead of passing a details we want to capture like `vaultAction`. For illustrative purposes, what do you think of something like:

```swift
func sendData(path: String, extraData: [String: Any], completion: @escaping (VaultResponse) -> Void, tracer: VaultProxyResponseTracer) {
    // code ...

    tracer.start() // calls DispatchTime.now() internally
    vgsCollect.sendData(path: path, extraData: mutableExtraData) { (response) in
        tracer.end() // calls DispatchTime.now() internally
        tracer.setPath(path)
        tracer.setMethod(HttpMethod.post)

        switch response {
        case .success(let code, let data, let urlResponse):
            // code ...
            tracer.setHttpStatusCode(code)
            tracer.captureAsLog()
        case .failure(let code, let data, let urlResponse, let error):
            // code ...
            tracer.setHttpStatusCode(code)
            tracer.captureAsLog()
        }
    }
}
```

devinmorgan (commented):
Nice work in this PR, Danilo! I wanted to run an opportunity past you to improve the maintainability/extensibility. Let me know what you think!

Sources/ForageSDK/Foundation/Telemetry/Metrics.swift
djoksimo: Discussed with Shardendu via slack DM

we've agreed on `response_time_ms` [instead of roundtrip or latency](https://docs.google.com/document/d/1ugFFT3Zs1q0I7NRQVdeJVnsYEIBGDtg37iM8PZz10PQ/edit?disco=AAAA4CIpuM8) across all of our vaulting metrics

Sources/ForageSDK/Foundation/Network/Collector.swift
djoksimo: 53be95834a2a44ea544bf379e72075d1138a3f1d

Sources/ForageSDK/Foundation/Telemetry/Metrics.swift
djoksimo: 53be95834a2a44ea544bf379e72075d1138a3f1d

devinmorgan (commented):
Thanks for considering upgrading the `struct` to be a class! I think there is one more opportunity to keep the `VaultProxyResponseMonitor.swift` internals out of the implementation details of `Collector.swift`. 

Since I don't have Xcode on my machine, I expressed this idea via Github comments which I admit is a bit awkward, so I appreciate your putting up with that.

The last opportunity that I see is basically to avoid passing details that the monitor/measurement cares about as params through `sendData` and instead only pass an instance of monitor/measurement as a param to `sendData` that is already configured with as much information as possible when passed as the arg. In this case, it seems like the monitor/measurement can be created already knowing its `vaultAction` since this is a decision made above the level of `Collector.swift`. In addition, we can spare `Collector.swift` from needing to be aware of the `enum VaultAction` and instead store that as a private implementation detail within `VaultProxyResponseMonitor.swift`. We can keep this detail private by using factory methods that initialize instances with the `vaultAction` already set.

If this makes sense and you find it agreeable, then great! If this doesn't make sense and you want to chat about it, I'm definitely down for that. And, if this doesn't make sense and you'd prefer just to ship this PR, that's OK as well 🙂

djoksimo:
> Thanks for considering upgrading the `struct` to be a class! I think there is one more opportunity to keep the `VaultProxyResponseMonitor.swift` internals out of the implementation details of `Collector.swift`.
> 
> Since I don't have Xcode on my machine, I expressed this idea via Github comments which I admit is a bit awkward, so I appreciate your putting up with that.
> 
> The last opportunity that I see is basically to avoid passing details that the monitor/measurement cares about as params through `sendData` and instead only pass an instance of monitor/measurement as a param to `sendData` that is already configured with as much information as possible when passed as the arg. In this case, it seems like the monitor/measurement can be created already knowing its `vaultAction` since this is a decision made above the level of `Collector.swift`. In addition, we can spare `Collector.swift` from needing to be aware of the `enum VaultAction` and instead store that as a private implementation detail within `VaultProxyResponseMonitor.swift`. We can keep this detail private by using factory methods that initialize instances with the `vaultAction` already set.
> 
> If this makes sense and you find it agreeable, then great! If this doesn't make sense and you want to chat about it, I'm definitely down for that. And, if this doesn't make sense and you'd prefer just to ship this PR, that's OK as well 🙂

Thank you for the thoughtful and thorough review @devinmorgan. I'm open to re-visiting this later and will ship now to expedite the release of these metrics.

I intentionally wanted `vaultAction` to be a parameter as I anticipated that it may be helpful for other purposes beyond performance metrics (ex: logging). 
I also wanted the performance measuring to live closely to the part we're monitoring (i.e. be handled by the `Collector`, rather than the `LiveForageService`).

devinmorgan:
> > Thanks for considering upgrading the `struct` to be a class! I think there is one more opportunity to keep the `VaultProxyResponseMonitor.swift` internals out of the implementation details of `Collector.swift`.
> > Since I don't have Xcode on my machine, I expressed this idea via Github comments which I admit is a bit awkward, so I appreciate your putting up with that.
> > The last opportunity that I see is basically to avoid passing details that the monitor/measurement cares about as params through `sendData` and instead only pass an instance of monitor/measurement as a param to `sendData` that is already configured with as much information as possible when passed as the arg. In this case, it seems like the monitor/measurement can be created already knowing its `vaultAction` since this is a decision made above the level of `Collector.swift`. In addition, we can spare `Collector.swift` from needing to be aware of the `enum VaultAction` and instead store that as a private implementation detail within `VaultProxyResponseMonitor.swift`. We can keep this detail private by using factory methods that initialize instances with the `vaultAction` already set.
> > If this makes sense and you find it agreeable, then great! If this doesn't make sense and you want to chat about it, I'm definitely down for that. And, if this doesn't make sense and you'd prefer just to ship this PR, that's OK as well 🙂
> 
> Thank you for the thoughtful and thorough review @devinmorgan. I'm open to re-visiting this later and will ship now to expedite the release of these metrics.
> 
> I intentionally wanted `vaultAction` to be a parameter as I anticipated that it may be helpful for other purposes beyond performance metrics (ex: logging). I also wanted the performance measuring to live closely to the part we're monitoring (i.e. be handled by the `Collector`, rather than the `LiveForageService`).

Revisiting sounds good, time to ship! 🚢

PR #68, opened 2023-08-09
[FX-398] iOS support dynamically updating the mask of ebt card

dleis612:
## What

Apply dynamic masking to the PAN based on expected card length.

Screen recording: https://joinforage.slack.com/archives/C031SJGSZV0/p1692122220569439

## Test Plan
Unit tests to validate the adjusted mask.

## How

can be released as-is. A later PR will make isValid default to true on empty state.

### DIFF
--- a/Sources/ForageSDK/Component/ForagePANTextField/MaskedUITextField.swift
+++ b/Sources/ForageSDK/Component/ForagePANTextField/MaskedUITextField.swift
@@ -92,6 +92,12 @@ internal class MaskedUITextField : UITextField, ObservableState {
             forageDelegate?.textFieldDidChange(self) // notify client about latest validation status.
         }
         
+        if isSpecialCard(newUnmaskedText) {
+            isValid = true
+            isComplete = newUnmaskedText.count >= 16 && newUnmaskedText.count <= 19
+            return
+        }
+        
         // Prior to seeing 6 digits, the PAN is considered valid but incomplete.
         if newUnmaskedText.count < 6 {
             isValid = true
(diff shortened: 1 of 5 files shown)


Sources/ForageSDK/Component/ForagePANTextField/MaskedUITextField.swift
```diff
@@ -0,0 +1,202 @@
...
+    
+    /// A delegate that informs the client about the state of the entered card number (validation, focus)
+    public weak var forageDelegate: ForageElementDelegate? {
+        didSet {
+            delegate = forageDelegate as? UITextFieldDelegate
+        }
+    }
+    
+    @IBInspectable private(set) public var isEmpty = true
+    // TODO: default to isValid = true on initialization
```
djoksimo, line 36: https://linear.app/joinforage/issue/FX-449/ios-update-default-isvalid-of-pin-and-pan-elements-to-be-true

djoksimo:
* main
  * **PR #68** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/68" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/68?utm_source=stack-comment).

Sources/ForageSDK/Component/ForagePANTextField/MaskedUITextField.swift
```diff
@@ -92,6 +92,12 @@ internal class MaskedUITextField : UITextField, ObservableState {
             forageDelegate?.textFieldDidChange(self) // notify client about latest validation status.
         }
         
+        if isSpecialCard(newUnmaskedText) {
```
dleis612, line 95: Only do this check in non-production.

Sources/ForageSDK/Component/ForagePANTextField/MaskedUITextField.swift
dleis612: Default this val to true :)

Tests/ForageSDKTests/MaskedUITextFieldTests.swift
```diff
@@ -20,15 +20,15 @@ final class MaskedUITextFieldTests: XCTestCase {
     
     override func tearDown() {
         maskedTextField = nil
+        ForageSDK.shared.environment = .sandbox
```
dleis612, line 23: Why is this set on teardown?

djoksimo: Because the `environment` is set to `.prod` in one of the tests. The `ForageSDK.setup` function doesn't update the environment value.

I've updated the tests here and added a comment: f92739a8ba22312b9f54e99e704d99738142a05a

```swift
    public class func setup(_ config: Config) {
        ForageSDK.config = config
       // we never update self.environment = ... here
    }
```

The fix is addressed in #75, which ensures the environment can be updated as desired (using the latest session token).

djoksimo (approved):

djoksimo (approved):

PR #69, opened 2023-08-09
[FX-415] iOS do not store pannumber in foragesdkshared

dleis612:
## What
Move the PAN value from shared storage to being owned by the individual PAN element itself.

## Test Plan
Adjust unit tests.

### DIFF
--- a/Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
+++ b/Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
@@ -23,6 +23,7 @@ protocol ForageSDKService: AnyObject {
         bearerToken: String,
         merchantAccount: String,
         customerID: String,
+        foragePanTextEdit: ForagePANTextField,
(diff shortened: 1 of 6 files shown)


Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
```diff
@@ -23,6 +23,7 @@ protocol ForageSDKService: AnyObject {
         bearerToken: String,
         merchantAccount: String,
         customerID: String,
+        foragePanTextEdit: ForagePANTextField,
```
djoksimo, line 26: can we update the Swift comments for this method definition to include 

```swift
/// - foragePanTextEdit: <helpful description> here
```

djoksimo (changes requested):
ty for fixing this!

We should create a TW ticket to update the public [readme](https://docs.joinforage.app/docs/ios) docs as well

djoksimo:
Supersedes this PR: https://github.com/teamforage/forage-ios-sdk/pull/86

PR #70, opened 2023-08-09
Add lint workflow

owenkim:
## What
Add a lint workflow for pull requests for baseline static analysis

### DIFF
--- a/.github/workflows/validations.yaml
+++ b/.github/workflows/validations.yaml
@@ -1,0 +1,14 @@
+name: PR Validations
+
+on: [pull_request]
+
+jobs:
+  pod-linting:
+    name: Pod Linting
+    runs-on: macOS-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      
+      - name: Pod Linting
+        run: pod lib lint --allow-warnings


.github/workflows/qa-tests.yaml
```diff
@@ -0,0 +1,45 @@
...
+          mkdir -p ~/Library/MobileDevice/Provisioning\ Profiles
+          cp $PP_PATH ~/Library/MobileDevice/Provisioning\ Profiles
+      - name: Build app
+        run: |
+          ls ~/Library/MobileDevice/Provisioning\ Profiles
+          cd Sample
+          mkdir -p ios_build
+          swift build
+          xcodebuild -scheme SampleForageSDK -destination 'generic/platform=iOS' CONFIGURATION_BUILD_DIR=./ios_build build DEVELOPMENT_TEAM=3D54TBBRF3 -allowProvisioningUpdates CODE_SIGNING_ALLOWED=NO
+          ls
```
devinmorgan, line 44: Clarifying question: does this job run a QA-Tests suite? I followed that it builds the SampleForageSDK but wasn't able to discern whether a test suite gets run. Thanks for any clarity!

devinmorgan (approved):
Not entirely sure how this PR causes the QA Tests to fail? 🤔 That aside, this linting check is super useful - thanks for taking the time improve our dev experience!

.github/workflows/qa-tests.yaml
dleis612: It will be doing that once it's done and working :) battling with iOS and github actions is not an easy task.

owenkim: I'm moving the QA test work to a separate PR so I will merge this to start the linting step.

PR #72, opened 2023-08-11
[FX-389] iOS add support for onfocuschanged event listener for bt

dleis612:
## What
Implement the focus event handler for BT based on changes they made for us.

## Test Plan
I tested visually in the sample app

### DIFF
--- a/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
+++ b/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
@@ -112,7 +112,7 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
                     self._isValid = elementEvent.valid
                     self._isComplete = elementEvent.complete
                 }
-                self.delegate?.textFieldDidChange(self)
+                self.delegate?.firstResponderDidChange(self)
             }
         }.store(in: &cancellables)
     }
--- a/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
+++ b/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
@@ -86,7 +86,7 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
             regexDigit
         ] as [Any]
         
-        let pinRegex = try! NSRegularExpression(pattern: "\\d{4}$")
+        let pinRegex = try! NSRegularExpression(pattern: "^\\d{4}$")
         
         try! textField.setConfig(options: TextElementOptions(mask: pinMask, validation: pinRegex))
         
--- a/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
+++ b/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
@@ -16,6 +16,8 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
     private enum BTEventType: String {
         case TEXT_CHANGE = "textChange"
         case MASK_CHANGE = "maskChange"
+        case BLUR = "blur"
+        case FOCUS = "focus"
     }
     
     private var _isEmpty: Bool = true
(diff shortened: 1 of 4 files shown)


Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
```diff
@@ -84,7 +86,7 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
             regexDigit
         ] as [Any]
         
-        let pinRegex = try! NSRegularExpression(pattern: "^\\d{4}$")
+        let pinRegex = try! NSRegularExpression(pattern: "\\d{4}$")
```
dleis612, line 89: I'm not entirely sure how this change in the regex didn't break the validation logic?? Makes me think BT's validation logic isn't working??

devinmorgan: I have no idea if the following note is helpful but figured I leave just in case:
- `^\\d{4}$` - has the `^`, which requires the match to happen at the start of the string.
  - `a2222` is not a match
  - `2222` is a match
- `\\d{4}$` - has no `^`, which allows the match to happen after the start of the string 
  - `a2222` is a match
  - `2222` is a match
Note that both of these regular expressions end in `$`, which means they require that the match ends at the end of the string
  - `2222a` is not a match for **either** of those regular expressions

So, if we are able to tell BT to have a max input length of 4, then both `^\\d{4}$` and `\\d{4}$` (as well as `\\d{4}`) will be equivalent since the start-of-line `^` and end-of-line `$` match characters are irrelevant since BT would already be enforcing that the sequence can be no longer than 4.

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
```diff
@@ -95,6 +97,22 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
...
                 self.delegate?.textFieldDidChange(self)
+            } else if (
+                elementEvent.type == BTEventType.FOCUS.rawValue ||
+                elementEvent.type == BTEventType.BLUR.rawValue
+            ) {
+                // Note: This is custom logic to unblock us while BT makes changes on their
+                // end. They had the incorrect initial state logic for their input field.
+                if (elementEvent.empty) {
+                    self._isEmpty = elementEvent.empty
+                    self._isValid = false
```
devinmorgan, line 108: Probably not too pressuring but in Android, [I had that `isEmpty = true` **is a valid state** but **not a complete state**](https://github.com/teamforage/forage-android-sdk/blob/b9fb8826b47b80f69b45a2eb04d0ce3dafd8d565/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt#L13). Does anybody have a strong opinion one way or the other? Just want to make sure the functionality is in sync across Android and iOS

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
```diff
@@ -95,6 +97,22 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
...
+                    self._isComplete = false
+                } else {
+                    self._isEmpty = elementEvent.empty
+                    self._isValid = elementEvent.valid
+                    self._isComplete = elementEvent.complete
+                }
+                self.delegate?.textFieldDidChange(self)
```
devinmorgan, line 115: Let me know if this is my lack of iOS understanding here but I wanted to clarify if we should be calling `self.delegate?.textFieldDidChange(self)` here or `self.delegate?.focusDidChange(self)`? It seems like this statement is living withing:
```swift
else if (
    elementEvent.type == BTEventType.FOCUS.rawValue ||
    elementEvent.type == BTEventType.BLUR.rawValue
) {
  // ...
  self.delegate?.textFieldDidChange(self) // should this be self.delegate?.focusDidChange(self) ?

}
```

devinmorgan (commented):
Nice work, Danny! Had a couple of clarifying questions on this PR and will approve right after!

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
dleis612: For the PIN, I `isValid` and `isComplete` both map to checking if the input is of length 4. If we change it, `isValid` would just be true at all times for the PIN right?

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
dleis612: Yikes you're right!

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
devinmorgan: I need to correct myself because what I said was misleading:
1. the default `isValid` state of an ForageElement is what I would advocate should be `true`
2. on any subsequent update, if the input is found to be empty, then `isValid = false`

In other words, what you have here is fine because I believe this function is the update step. However, a little bit above in this file `isValid` is initialized to false (link [here](https://github.com/teamforage/forage-ios-sdk/blob/8bb9d898da28881940b2031e05e3916507622630/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift#L26)). I would advocate that it should be initialized to `true` so that any merchant does not kick things off by displaying an error message of their own because the default state of the `ForageElement` suggest the field is invalid.

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
devinmorgan: Ah OK - glad I wasn't going crazy there! I saw that you fixed this in [`fdbadf8`](https://github.com/teamforage/forage-ios-sdk/pull/72/commits/fdbadf83f75b8da6e698d9f084da27595e764221). Will re-view now!

devinmorgan (approved):
Thanks for considering my questions! Nice work!

djoksimo (approved):

PR #78, opened 2023-08-21
[SYM-106] Create floating placeholder for PAN textfield.

milosbswe:
## What

- Create floating placeholder for PAN textfield.

Screenshot 2023-08-21 at 23 14 06

## Test Plan


Sources/ForageSDK/Component/ForagePANTextField/FloatingTextField.swift
```diff
@@ -0,0 +1,314 @@
...
+                DispatchQueue.main.async {
+                    self.layoutIfNeeded()
+                }
+            }
+        } else {
+            animations()
+        }
+    }
+    
+    private func hideFlotingLabel(_ animated: Bool) {
```
dleis612, line 159: Floating*

dleis612 (changes requested):
This looks great! Could add a small amount of padding on the top and bottom of the input field to push the text a little further from the border? Also, is it possible to use a different "clearButton"? Our customers want our design to match the mockup as closely as possible.

Sources/ForageSDK/Component/ForagePANTextField/FloatingTextField.swift
milosbswe: Typo... I will fix it. Thanks.

dleis612 (approved):

devinmorgan (commented):
@dleis612, feel free to say this is not a priority right now, in which case, @milosbswe, feel free to ignore my comments here:

Hey Milos! Thanks for the good work here. In the near future, we would like to add a GitHub check that enforces at least 80% code coverage for every PR, and this PR adds ~400 LOC with no test coverage. 

From what I can tell, all of the logic/calculations in this PR appear to live directly in the view layer. I am less familiar with how easy or difficult it is to test view-layer logic in iOS. However, if it is easy enough to write unit tests on `FloatingTextField` directly, could you add some tests of the pure calculations on the view layer? However, if it's more complex to test `FloatingTextField` directly, can you extract some of this class's calculations into pure functions with associated unit tests? Here are some of the places that I noticed had untested calculations:

vars

- `floatLabelWidth`

methods

- `setFloatLabelAlignment`
- `insetRectForEmptyBounds`
- `insetRectForBounds`
- `rightViewRect`
- `clearButtonRect`

dleis612:
> @dleis612, feel free to say this is not a priority right now, in which case, @milosbswe, feel free to ignore my comments here:
> 
> Hey Milos! Thanks for the good work here! In the near future, we would like to add [a GitHub check that enforces at least 80% code coverage for every PR](https://github.com/teamforage/forage-ios-sdk/actions/runs/5951241567/job/16140671736?pr=78#step:9:14), and this PR adds ~400 LOC with no test coverage.
> 
> From what I can tell, all of the logic/calculations in this PR appear to live directly in the view layer. I need to become more familiar with how easy or difficult it is to test view-layer logic in iOS. If it is easy enough to write unit tests on `FloatingTextField` directly, could you add some tests of the pure calculations on the view layer? However, if it's more complex to test `FloatingTextField` directly, can you extract some of this class's calculations into pure functions with associated unit tests? Here are some of the places that I noticed had untested calculations:
> 
> vars
> 
> * `floatLabelWidth`
> 
> methods
> 
> * `setFloatLabelAlignment`
> * `insetRectForEmptyBounds`
> * `insetRectForBounds`
> * `rightViewRect`
> * `clearButtonRect`

I agree @devinmorgan! @milosbswe this would be very helpful.

Tests/ForageSDKTests/FloatingTextFieldTests.swift
```diff
@@ -0,0 +1,238 @@
...
+        let topInset = paddingYFloatLabel + floatingTextField.bounds.size.height
+        let textOriginalY = floatingTextField.frame.height / 2.0
+        var textY = topInset - textOriginalY
+        
+        if textY < 0 { textY = topInset }
+        
+        let expectedRect = CGRect(x: x,
+                                  y: ceil(textY) - 2,
+                                  width: bounds.size.width - x - paddingX,
+                                  height: bounds.height - topInset)
```
devinmorgan, line 115: This code appears copy/pasted from `private func insetRectForBounds` (inside of `FloatingTextField.swift`). It's important to remember that writing tests should meaningfully increase our confidence that the original code works as expected. If a test copies the implementation verbatim, then we are only gaining confidence that the Swift compiler works as expected; we are not increasing our confidence that the code we wrote works as we expect.

The way that we gain confidence that the code we wrote works as expected is by documenting very simple cases inside of unit tests that other developers will easily be able to understand. Here is an example (for illustration purposes only) of a test that is easy to read and that tests a single case. Noteably it does not copy the implementation details of the test and does not include if-statements or other calculations.
```swift
func test_editingRect_forGivenBounds() {
    // Setup
    let bounds = CGRect(x: 0, y: 0, width: 100, height: 50) // A simple 100x50 rect for clarity.

    // Let's assume that the FloatingTextField is designed such that when given these bounds and 
    // no text is set, the editing rect should be inset by 10 points from each side.
    let expectedRect = CGRect(x: 10, y: 10, width: 80, height: 30)

    // Act
    let resultRect = floatingTextField.editingRect(forBounds: bounds)

    // Assert
    XCTAssertEqual(resultRect, expectedRect)
}

```

devinmorgan (changes requested):
Hey Milos, I appreciate you taking the time to add these tests and getting the coverage above 80%. I flagged a few concerns that I had with the tests. Do these concerns make sense to you? Feel free to Slack me if you have any questions. If my concerns sound unclear, then don't worry about it, and I'll approve the PR tomorrow morning. Thanks again for your work on this!

devinmorgan (approved):

Tests/ForageSDKTests/FloatingTextFieldTests.swift
milosbswe: Updated. Thanks

PR #79, opened 2023-08-22
[FX-470] Replace Datadog v1 package with v2 "vendor/" approach

djoksimo:
## What

- Fixes Swift dependency coupling for Datadog package.
- Add "Vendor/" directory with required DatadogCore and DatadogLogs files from dd-sdk-ios v2.1.0
- Add needed "DatadogPrivate-Objc" code and bridge objective C code with our Swift package
- Update implementation of Datadog logging (found in `DatadogLogger` class in `ForageLogger.swift`) to use Datadog v2
- Ensures logs are disabled in unit tests

### TODO

- ~~ideally, include Datadog tests in our code and run them alongside our unit tests – to ensure Datadog files were copied over reliably.~~ Unit tests omitted for the reasons described [in this thread](https://joinforage.slack.com/archives/C056QGRF6S3/p1692827519183079?thread_ts=1692714236.892159&cid=C056QGRF6S3)

## Why

- More info in [this design doc](https://docs.google.com/document/d/1BU609qv7qSDN-tdNaJP-sAyrgeM6REe9tvcMk5CxN3c/edit#heading=h.cix334vkod5s)
- See [SDK logging design doc](https://docs.google.com/document/d/1NYe6JtnBJAVrtjIKREdwoBt3lqh18vfXqzj9EaodXhI/edit#heading=h.o0dx96s3q4m) for general logging principles + goals
- See [Logger SDK interface proposal](https://www.notion.so/Mobile-SDK-Logger-Interface-Proposal-f97d859471a0424c83d19e46665aa1f9)

## Test Plan

- ✅ Unit Tests passed locally

- [x] [logs appear in Datadog](https://app.datadoghq.com/logs?query=service%3Aios-sdk%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1692675202687&to_ts=1692676821497&live=false) as desired (see #55 for more info on what kind of logs are expected)

## How

Can be released as-is


djoksimo:
* main
  * **PR #54** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/54" target="_blank">Graphite</a> 
    * **PR #79** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/79" target="_blank">Graphite</a>  👈
      * **PR #65** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/65" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/79?utm_source=stack-comment).

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
```diff
@@ -145,6 +145,8 @@ extension ForageSDK: ForageSDKService {
                 ))
                 .notice("Called ForageSDK.shared.capturePayment for Payment \(paymentReference)", attributes: nil)
             
+            foragePinTextEdit.disable()
```
dleis612, line 148: Why did you move these `disable` calls below the log calls?

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -1,11 +1,17 @@
...
+//  ForageLogger.swift
 //
 //
 //  Created by Danilo Joksimovic on 2023-07-26.
 //
 
-import Datadog
+// NOTE: This logger is built upon a "stripped" version of the Datadog package v2.1.0, found in the Sources/ForageSDK/Vendor directory.
+// "Stripped" refers to the inclusion of only the DatadogCore and DatadogLogs modules, which are essential for emitting logs as described at https://docs.datadoghq.com/logs/log_collection/ios/.
+// Initially, we decided against forking the Datadog package due to the complexity of resolving name clashes and other integration challenges.
```
dleis612, line 10: Since our packages are public, can we maybe put these into a notion document and link the document here? That way, those outside the organization can't see this info.

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
djoksimo: We want the span/trace to start as soon as the method is called.

e.g. timestamp of "Called ForageSDK.shared.capturePayment" matches the same moment "capturePayment" was called.

Suppose "disable" was called first and it took a while to complete executing or threw some kind of error, we would have inaccurate timestamps and/or the "capturePayment" would never get logged.

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -69,29 +68,42 @@ internal protocol ForageLogger {
...
+
+    private var logger: LoggerProtocol? = nil
     private var config: ForageLoggerConfig? = nil
-        
+    
     required internal init(_ config: ForageLoggerConfig? = ForageLoggerConfig()) {
+        if isUnitTesting() {
+            // avoid emitting logs when running unit tests
+            return
+        }
```
devinmorgan, line 84: A suggestion for potential improvement at a better time:

The use of `isUnitTesting()` gets the job done and allows us to ship quickly. I think it might be beneficial for us to consider having a more flexible logging mechanism as we continue to scale our application. We could have multiple types of loggers that implement the same interface. For example:

- `DatadogLogger`: For our production and staging logs (already implemented)
- `LocalLogger`: This might be used to log statements locally, especially useful during development.
`NothingLogger`: A logger that essentially does nothing, which can be useful in certain scenarios like unit testing.

Here's a simple example of how this could look inside the ForageSDK:

```swift
// very much abbreviated
public class ForageSDK {
    private init() {
	let useProdLogger = !isUnitTesting()
	let logger = useProdLogger ? DatadogLogger() : NothingLogger() 
    }
}
```
This way, when we're running unit tests, we wouldn't even initialize the DatadogLogger and can avoid adding branching logic to our Loggers since each one would only focus on one use case (e.g. for prod, unit testing etc)

I created [this ticket](https://linear.app/joinforage/issue/FX-490/ios-create-multiple-instances-of-loggers-to-avoid-context-based-if) so we can discuss on Linear and consider it (or not) in the future.

dleis612 (approved):

.slather.yml
```diff
@@ -5,4 +5,9 @@ ignore:
   - ../Library/*
   - ../../Library/*
   - ../../../Library/*
+  - "**/Vendor/DatadogCore/*"
+  - "**/Vendor/DatadogInternal/*"
+  - "**/Vendor/DatadogLogs/*"
+  - "DatadogPrivate-Objc/*"
+  - "**/DatadogPrivate-Objc/*"
   - Tests/*
```
devinmorgan, line 13: I find it slightly ironic that we ignore the `Tests/` directory for coverage. Am I missing something? I know its not part of this PR but noticed anyhow

devinmorgan (approved):
I looked over all of the files that were not under the `/Vendor` directory. Since the logs appear in datadog as we hope, this sounds good to me.

.slather.yml
dleis612: If you didn't include these, we would need to write tests to cover our tests...

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
djoksimo: Agreed! Would definitely be good to have a factory or something where we can have more control over logs getting emitted to the console, the verbosity and so on.

.slather.yml
djoksimo: We don't test our unit tests so we omit these from the coverage check

devinmorgan: OOOOOO that makese sense. For some reason I thought it had to do with don't run any of the tests under `Test/` for coverage purposes. Got it - we definitely don't want to write tests for our tests 😂. Thanks for the clarity all!

PR #81, opened 2023-08-23
[FX-486] PAN uses num pad, PINs use phone pad

djoksimo:
## What

PAN (no letters)

PIN inputs use telephone pad (with letters)

**Remark: when I tested this on iOS 14.0.1 against an iPhone 8., the phone pad still appeared for the PAN field – but this feature seems to work on newer iPhones (tested with iPhone 14 using iOS 16.2)**

## Why

Better UX/CX

## Test Plan

## How

Can be released as-is


djoksimo:
* main
  * **PR #81** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/81" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/81?utm_source=stack-comment).

dleis612 (approved):

PR #88, opened 2023-08-30
[FX-512] Add ForageSDK.swift unit tests

djoksimo:
## What

- Add unit tests for `ForageSDK.swift`
- Remove `ForageSDK.shared.service = nil` statements [FX-416](https://linear.app/joinforage/issue/FX-416/ios-remove-foragesdksharedservice-=-nil-statements-from-tests)

## Why

- helps ensure the rest of the changes in this PR stack work and reduce the risk of regressions
- helps ensure logger is initialized, which is not covered by our E2E tests.

## Test Plan

- ✅ iOS QA Tests passed locally
- ✅ Unit Tests passed locally

## How

Can be released as-is, with the rest of the PRs in this stack.

PR #93, opened 2023-09-05
[FX-525] Fix environment not set properly

djoksimo:
## What

- Ensure environment, merchantID and sessionToken are set as desired.

### Consequences

- The client will not be able to conveniently + programmatically update the `sessionToken` at runtime (i.e. session token refresh). AFAIK, this functionality is not officially supported
- The client will not be able to programmatically update the `merchantID` at runtime. This functionality is not necessary for GoPuff, but will be for future tenants.

### The issue

Fixed an initialization issue in `ForageSDK` where the shared instance was being accessed within `ForageSDK.setup(...)`, causing it to initialize before the environment was set."

This in turn meant that Datadog and LaunchDarkly received the wrong environment.

### Additional context

- we throw a fatal assertion error if the client tries to access `ForageSDK.shared` before calling `ForageSDK.setup`.

### Why tests were removed 

I added tests in #88; however, because `ForageSDK` is a singleton that is shared across our unit tests. We cannot write unit tests for `ForageSDK.init`.

#### Next steps

I've [created a ticket](https://linear.app/joinforage/issue/FX-526/spike-improve-foragesdkswift) to properly refactor `ForageSDK.swift` – as it's currently brittle, untestable via unit tests and prone to bugs (as illustrated by this bugfix PR)

## Test Plan

- ✅ Unit Tests passed locally

## How

Can be released as-is.


djoksimo:
* main
  * **PR #93** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/93" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/93?utm_source=stack-comment).

devinmorgan (approved):
Singletons give me the shivers at this point. Nice work!

PR #94, opened 2023-09-05
[FX-529] Add support for error details

djoksimo:
## What

- Add support for `ForageErrorObj.details` 
- Add helpful comments to error structs, to improve dev experience and reduce integration time
- Show details in payment capture page
- Reduce duplication in sample app payment capture page

## Why

https://linear.app/joinforage/issue/FX-529/ios-expose-details-field-for-forageerror

## Test Plan

- ✅ Unit Tests passed locally

Used this [test card](https://docs.joinforage.app/docs/test-ebt-cards#:~:text=Expired%20card-,4444444444444451,-Insufficient%20funds) to force insufficient funds error

Created [this ticket](https://linear.app/joinforage/issue/FX-532/ios-add-e2e-test-assertion-for-error-details-on-insufficient-funds) to add e2e tests for covering this change.

## How

Can be released as-is.

Can share the following snippet with GoPuff for error-handling info:

```swift
 case .failure(let error):
    if let forageError = error as? ForageError? {
        let firstError = forageError?.errors.first
        
        if (firstError?.code == "ebt_error_51") {
            let snapBalance = firstError?.details?.snapBalance ?? "N/A"
            let cashBalance = firstError?.details?.cashBalance ?? "N/A"

            // display these details
            print("firstForageError.details: remaining balances are SNAP: \(snapBalance), EBT Cash: \(cashBalance)")
        }
    }
```

### DIFF
--- a/Sources/ForageSDK/Foundation/Network/Model/ForageErrorModel.swift
+++ b/Sources/ForageSDK/Foundation/Network/Model/ForageErrorModel.swift
@@ -23,12 +23,50 @@ internal struct ForageErrorSource: Codable {
     let ref: String
 }
 
+/// Represents an error that occurs when a request to submit a `ForageElement` to the Forage API fails.
 public struct ForageError: Error, Codable {
+    /// An array of error objects returned from the Forage API.
     public let errors: [ForageErrorObj]
 }
 
+/// Contains additional details about a Forage API error.
+public struct ForageErrorDetails: Codable {
+    /// The remaining SNAP balance
+    /// Only populated when the error code is [ebt_error_51](https://docs.joinforage.app/reference/errors#ebt_error_51)
+    public let snapBalance: String?
+    
+    /// The remaining EBT Cash balance.
+    /// Only populated when the error code is [ebt_error_51](https://docs.joinforage.app/reference/errors#ebt_error_51)
+    public let cashBalance: String?
+    
+    private enum CodingKeys : String, CodingKey {
+        case snapBalance = "snap_balance"
+        case cashBalance = "cash_balance"
+    }
+}
+
+/// Represents a detailed error object returned by the Forage API.
+/// Provides additional context about the HTTP status, error code, and developer-facing message.
+/// [Learn more about SDK errors](https://docs.joinforage.app/reference/errors#sdk-errors)
 public struct ForageErrorObj: Codable {
+    /// The HTTP status that the Forage API returns in response to the request.
     public let httpStatusCode: Int
+    
+    /// A short string explaining why the request failed. The error code string corresponds to the HTTP status code.
+    /// [Learn more about SDK error codes](https://docs.joinforage.app/reference/errors#code-and-message-pairs-1)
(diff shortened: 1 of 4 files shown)


djoksimo:
* main
  * **PR #94** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/94" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/94?utm_source=stack-comment).

devinmorgan (approved):
This works! 🚢

dleis612 (approved):

Sources/ForageSDK/Foundation/Network/Model/ForageErrorModel.swift
```diff
@@ -23,12 +23,50 @@ internal struct ForageErrorSource: Codable {
...
+    
+    /// The remaining EBT Cash balance.
+    /// Only populated when the error code is [ebt_error_51](https://docs.joinforage.app/reference/errors#ebt_error_51)
+    public let cashBalance: String?
+    
+    private enum CodingKeys : String, CodingKey {
+        case snapBalance = "snap_balance"
+        case cashBalance = "cash_balance"
+    }
+}
```
devinmorgan, line 46: Hey Danilo, I just realized this in the Android code for this fix, but I wanted to flag a possible issue regarding type safety:

Right now, with `ForageErrorDetails` directly having `snapBalance` and `cashBalance` as fields, the swift compiler won't enforce that developer performs an if-check to confirm that `forage_code == "ebr_error_51"`. So, even though you make that recommendation clear in your notes, a developer could accidentally not include the if-check. 

```swift
 if (firstError?.code == "ebt_error_51") { // compiler won't force this since our types say that if its non-null then it has `snapBalance` / `cashBalance`
    let snapBalance = firstError?.details?.snapBalance ?? "N/A"
    let cashBalance = firstError?.details?.cashBalance ?? "N/A"

    // display these details
    print("firstForageError.details: remaining balances are SNAP: \(snapBalance), EBT Cash: \(cashBalance)")
}
```

If instead you have `ForageErrorDetails` referring to enum and that `EbtError51` is one possible value, then the compiler would enforce that folks have to do an if-check before trying to access `snapBalance` and `cashBalance`

djoksimo: For context, our other SDKs (iOS, Android, JS) and [Stripe](https://github.com/stripe/stripe-js/blob/1cfce6fbf0edec4c683d755badd6fe7b153e11ec/types/stripe-js/stripe.d.ts#L1429) use string > enums for the error codes. That being said, it sounds like we should continue to treat them as strings rather than enums because the SDKs are versioned and error codes could dynamically change/evolve or be dictated by upstream processors (like FIS).

The client will access `snapBalance` or `cashBalance` from the error details when the insufficient funds error occurs.

PR #95, opened 2023-09-06
[FX-384] Add trace ID to iOS SDK

dleis612:
## What
Add a trace ID to the iOS SDK so that we can group logs together.

## Test Plan
Tested the functionality and validated in datadog.
Screenshot 2023-09-08 at 11 01 42 AM

### DIFF
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -17,6 +17,7 @@ public class ForageSDK {
     internal var logger: ForageLogger? = nil
     internal var merchantID: String = ""
     internal var sessionToken: String = ""
+    internal var traceId: String = ""
     
     public var environment: Environment = .sandbox
     // Don't update! Only updated when releasing.
@@ -33,14 +34,15 @@ public class ForageSDK {
         self.environment = Environment(sessionToken: config.sessionToken)
         self.merchantID = config.merchantID
         self.sessionToken = config.sessionToken
-        
+        self.traceId = ForageSDK.generateTraceId()
         // ForageSDK.shared.environment is not set
         // until the end of this initialization
         // so we have to provide the environment from the config
         let logger = DatadogLogger(
             ForageLoggerConfig(
                 environment: self.environment,
-                prefix: ""
+                prefix: "",
+                traceId: traceId
             )
         )
         self.logger = logger
@@ -49,6 +51,16 @@ public class ForageSDK {
         VGSCollectLogger.shared.disableAllLoggers()
         let provider = Provider(logger: logger)
         self.service = LiveForageService(provider: provider, logger: logger)
+    }
+    
+    private static func generateTraceId() -> String {
+        var traceSeed = ""
+        for _ in 0..<14 {
+            let randomDigit = UInt64(arc4random_uniform(10))
+            traceSeed = traceSeed + String(randomDigit)
+        }
+        return "33" + traceSeed
+        
     }
     
     /**
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -2,8 +2,6 @@
 //  ForageSDK.swift
 //  ForageSDK
 //
-//  Created by Symphony on 18/10/22.
-//
 
 import Foundation
 import VGSCollectSDK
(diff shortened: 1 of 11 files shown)


Sources/ForageSDK/ForageSDK.swift
```diff
@@ -51,6 +53,16 @@ public class ForageSDK {
...
     }
     
+    private static func generateTraceId() -> String {
+        var traceSeed = ""
+        for _ in 0..<14 {
+            let randomDigit = UInt64(arc4random_uniform(10))
+            traceSeed = traceSeed + String(randomDigit)
+        }
+        return "33" + traceSeed
+
```
djoksimo, line 63: can remove this trailing newline

Sources/ForageSDK/Foundation/Network/ForageAPI.swift
```diff
@@ -60,7 +60,8 @@ extension ForageAPI: ServiceProtocol {
                 "authorization": "Bearer \(model.authorization)",
                 "content-type": "application/json",
                 "accept": "application/json",
-                "API-VERSION": "2023-05-15"
+                "API-VERSION": "2023-05-15",
+                "x-datadog-trace-id": model.traceId ?? ""
```
djoksimo, line 64: should we set this header on all HTTP requests with one line, rather than drilling it through every request?

```swift
// in Provider.swift or ServiceProtocol.swift
            request.setValue(traceId, forHTTPHeaderField: "x-datadog-trace-id")
```

Sources/ForageSDK/Foundation/Network/ForageAPI.swift
```diff
@@ -12,10 +12,10 @@ import Foundation
  */
 enum ForageAPI {
     case tokenizeNumber(request: ForagePANRequestModel)
```
djoksimo, line 14: should `tokenizeNumber` also use a traceId?

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -116,6 +119,8 @@ internal class DatadogLogger : ForageLogger {
             in: datadogInstance
         )
         
+        self.logger?.addAttribute(forKey: "trace_id", value: self.config?.traceId)
```
djoksimo, line 122: we can instead pass `traceId` via the `ForageLoggerContext` object, which will set it when we call addContext [here](https://github.com/teamforage/forage-ios-sdk/pull/95/files#diff-e079539e1d6fdb51dad7a4f72feeb7c0b00815118a211e5a5480af80314a9732R124)

```swift
// in ForageSDK.swift

// in ForageSDK.swift
   let logger = DatadogLogger(
            ForageLoggerConfig(
                environment: self.environment,
                prefix: "",
                context: ForageLogContext(
                    traceId: generateTraceId()
                )
            )
        )
```

So we can safely 
- remove the new line you've added on 122.
- remove `traceid` as a param in the ctor of `ForageLogger
- remove `traceId` as a member of ForageLogger

djoksimo (changes requested):

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -78,17 +78,24 @@ internal protocol ForageLogger {
     /// - Parameter error: An optional error object associated with the critical log.
     /// - Parameter attributes: An optional dictionary of key-value pairs providing additional context for the log.
     func critical(_ message: String, error: Error?, attributes: [String: Encodable]?)
+    
+    /// The ForageSDK object needs a handler to the TraceId in order to pass it along to each endpoint call.
+    /// - Returns: The traceID that was generated at the beginning of the app's lifecycle.
+    func getTraceId() -> String
```
djoksimo, line 84: Nit: should be refactored to `getTraceID` for swift conventions; but feel free to ignore this, since this code is internal

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -78,17 +78,24 @@ internal protocol ForageLogger {
...
 }
 
 /// Read this [document](https://docs.google.com/document/d/1BU609qv7qSDN-tdNaJP-sAyrgeM6REe9tvcMk5CxN3c/edit#bookmark=id.w44b8kk07o7o) for to learn more about how DatadogLogger works
 internal class DatadogLogger : ForageLogger {
     private static let DD_CLIENT_TOKEN: String = "pub1e4572ba0f5e53df108c333d5ec66c02"
     private static let DD_SERVICE_NAME: String = "ios-sdk"
     private static let DD_SDK_INSTANCE_NAME: String = "forage"
+    
+    // DO NOT UPDATE! Generate 1 TraceID per living session of the app
+    internal static var traceId: String = generateTraceId()
```
djoksimo, line 94:
```suggestion
    // DO NOT UPDATE! Generate 1 TraceID per living session of the app
    internal static let traceId: String = generateTraceId()
```

we can keep the comment, but this can programmatically enforce this

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -200,4 +215,13 @@ internal class DatadogLogger : ForageLogger {
         }
         return message
     }
+    
+    private static func generateTraceId() -> String {
```
djoksimo, line 219:
```suggestion
    private static func generateTraceID() -> String {
```

can ignore this^

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
```diff
@@ -116,6 +123,10 @@ internal class DatadogLogger : ForageLogger {
             in: datadogInstance
         )
         
+        // Do this explicitly in the constructor so that we confirm each logger that is created
+        // has the trace_id field.
+        self.logger?.addAttribute(forKey: "trace_id", value: DatadogLogger.traceId)
```
djoksimo, line 128: should we add a line like this to ensure the Trace ID is visible in the attributes/context section of a log?

```swift
        self.logger?.addAttribute(forKey: "forage_trace_id", value: DatadogLogger.traceId)
```

djoksimo (approved):
left a few comments, but can be merged as-is!

Sources/ForageSDK/Foundation/Telemetry/ForageLogger.swift
dleis612: I think it's fine without it, it's appended to the each log statement already. It just appears under the trace tab rather than as an attribute.

djoksimo (approved):

PR #106, opened 2023-09-12
[POP-180] Adding jitter to retry logic

nitinsampath:
## What
Adding +/- 25ms jitter to polling so that we can avoid a thundering herd issue if our clients retry all at the same time. Now all retries to polling will have some randomness added to their delay interval.

## Test Plan

- ✅ | ❌ Not a front-end change
- ✅ | ❌ Unit Tests passed locally

### DIFF
--- a/Sources/ForageSDK/Foundation/Network/LDManager.swift
+++ b/Sources/ForageSDK/Foundation/Network/LDManager.swift
@@ -11,12 +11,79 @@ import LaunchDarkly
 
 internal protocol LDClientProtocol {
     func doubleVariationWrapper(forKey key: String, defaultValue: Double) -> Double
+    func jsonVariationWrapper(forKey key: String, defaultValue: LDValue) -> LDValue
 }
 
 extension LDClient: LDClientProtocol {
     func doubleVariationWrapper(forKey key: String, defaultValue: Double) -> Double {
         return self.doubleVariation(forKey: key, defaultValue: defaultValue)
     }
+    
+    func jsonVariationWrapper(forKey key: String, defaultValue: LDValue) -> LDValue {
+        return self.jsonVariation(forKey: key, defaultValue: defaultValue)
+    }
+}
+
+extension LDValue {
+    
+    /// Converts the original LDValue returned from the polling feature flag to an Array of Ints.
+    internal func convertToIntArray() -> [Int] {
+        let intervalArrayAsLDValue = self.extractArray(pollingFlagObj: self)
+        let intervalArrayAsArray = self.convertToArray(pollingArray: intervalArrayAsLDValue)
+        return self.convertArrayValsToInts(arrayOfInts: intervalArrayAsArray)
+    }
+    
+    /// Inspects the original LDValue returned from the polling feature flag and extracts the value mapped to INTERVALS
+    ///
+    /// - Parameters:
+    ///   - pollingFlagObj: The `LDValue` representing a JSON object.
+    internal func extractArray(pollingFlagObj: LDValue) -> LDValue {
+        let defaultLDValueArray: LDValue = LDValue(arrayLiteral: [])
+        switch pollingFlagObj {
+        case .object(let dictValue):
+            for (key, value) in dictValue {
+                if (key == INTERVALS) {
+                    return value
+                }
+            }
+            return defaultLDValueArray
+        default:
+            return defaultLDValueArray
+        }
+    }
+    
+    /// Converts the LDValue to an Array.
+    ///
+    /// - Parameters:
+    ///   - pollingArray: The `LDValue` representing an Array.
+    internal func convertToArray(pollingArray: LDValue) -> [LDValue] {
+        switch pollingArray {
+        case .array(let intervalVals):
+            return intervalVals
+        default:
+            return []
+        }
+    }
+
+    /// Converts an Array of LDValue to an Array of Ints.
+    ///
+    /// - Parameters:
+    ///   - arrayOfInts: The `LDValue` representing an Array of Ints.
+    internal func convertArrayValsToInts(arrayOfInts: [LDValue]) -> [Int] {
+        return arrayOfInts.map { timeInMs in
+            switch timeInMs {
+            case .number(let valAsNum):
+                return Int(valAsNum)
+            default:
+                return 1000
+            }
+        }
+    }
+}
+
+internal protocol LDManagerProtocol {
+    func getVaultType(ldClient: LDClientProtocol?, genRandomDouble: () -> Double, fromCache: Bool) -> VaultType
+    func getPollingIntervals(ldClient: LDClientProtocol?) -> [Int]
 }
 
 /**
@@ -43,7 +110,18 @@ public enum VaultType: String {
  */
 private enum FlagType: String {
     case vaultPrimaryTrafficPercentage = "vault-primary-traffic-percentage"
+    case isoPollingWaitIntervals = "iso-polling-wait-intervals"
 }
+
+/**
+ Key value used to grab the dictionary returned by the "iso-polling-wait-intervals" flag
+ */
+private let INTERVALS = "intervals"
+
+/**
+ LaunchDarkly prefix
+ */
+private let LAUNCH_DARKLY_PREFIX = "LaunchDarkly"
 
 /**
  Contexts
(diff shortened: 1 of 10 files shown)


Sources/ForageSDK/Foundation/Network/LiveForageService.swift
```diff
@@ -310,14 +310,23 @@ extension LiveForageService: Polling {
...
+    }
     
     /// Support function to update retry count and interval between attempts.
     ///
     /// - Parameters:
     ///  - completion: Which will return after an wait.
     private func waitNextAttempt(completion: @escaping () -> ()) {
         retryCount = retryCount + 1
-        DispatchQueue.main.asyncAfter(deadline: .now() + self.intervalBetweenAttempts) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + self.intervalBetweenAttempts + self.getJitterAmount()) {
```
djoksimo, line 329: can we separate this sum into a separate/new variable for readability

Sources/ForageSDK/Foundation/Network/LiveForageService.swift
```diff
@@ -310,14 +310,23 @@ extension LiveForageService: Polling {
...
     }
+
+    /// We generate a random jitter amount to add to our retry delay when polling for the status of
+    /// Payments and Payment Methods so that we can avoid a thundering herd scenario in which there are
+    /// several requests retrying at the same exact time.
+    /// 
+    /// Returns a random double between -25 and 25  
+    internal func getJitterAmount() -> Double {
+        return Double(Int.random(in: -25...25)) / 1000.0
+    }
```
djoksimo, line 321: I think we can use the 2nd `using` / `rng` param of `Int.random` to inject the mock random number generator in our unit test.

and default to the [SystemRandomNumberGenerator](https://developer.apple.com/documentation/swift/systemrandomnumbergenerator) when it's not provided (i.e. called in the LiveForageService.

maybe something like this

```swift
protocol RandomNumberGeneratorProtocol {
    mutating func next() -> UInt64
}

internal func getJitterAmount(randomNumberGenerator: RandomNumberGeneratorProtocol? = nil) -> Double {
    let rng = randomNumberGenerator ?? SystemRandomNumberGenerator() // fallback to system one
    return Double(Int.random(in: -25...25, using: &rng)) / 1000.0
}
```

In the tests, maybe something like this:

```swift
struct FixedSeedRandomNumberGenerator: RandomNumberGeneratorProtocol {
    mutating func next() -> UInt64 {
        return <insert-fixed-seed-here>
    }
}

// in test case:
getJitterAmount(randomNumberGenerator: FixedSeedRandomNumberGenerator())
```

djoksimo: If this takes a while to figure out. I say we just add some print statements showcasing that this works as intended and merge this PR.

djoksimo (changes requested):

Sources/ForageSDK/Foundation/Network/LiveForageService.swift
nitinsampath: Having some issues getting this to work, I'll go with my original approach for now, but we can discuss later further if I was perhaps missing something

nitinsampath: I'm not sure why the coverage is failing though, we can also discuss this when you get the chance

Tests/ForageSDKTests/ForageServiceTests.swift
```diff
@@ -205,6 +205,42 @@ final class ForageServiceTests: XCTestCase {
...
+        let mockSession = URLSessionMock()
+        mockSession.data = forageMocks.capturePaymentSuccess
+        mockSession.response = forageMocks.mockSuccessResponse
+        let service = LiveForageService(provider: Provider(mockSession))
+        
+        let jitterAmount = service.getJitterAmount()
+        
+        // Assert
+        XCTAssertNotNil(jitterAmount, "Jitter amount should not be nil")
+        XCTAssertTrue(jitterAmount >= -0.025 && jitterAmount <= 0.025, "Jitter amount should be within -0.025 and 0.025")
```
djoksimo, line 219: Normally, we should avoid relying on random values in our unit tests to [keep them deterministic](https://www.browserstack.com/guide/unit-testing-best-practices#:~:text=Ensure%20tests%20are%20deterministic).

Given this is harder to set up and mock in Swift, we can keep the test as-is for now and create a ticket to refactor `getJitterAmount` to be mockable (fix the RNG seed value).

Sources/ForageSDK/Foundation/Network/LiveForageService.swift
```diff
@@ -310,14 +310,24 @@ extension LiveForageService: Polling {
...
```
djoksimo, line 321: @dleis612 

```suggestion
    /// Returns a random double between -25 ms and 25 ms
    internal func jitterAmountInSeconds() -> Double {
        return Double(Int.random(in: -25...25)) / 1000.0
    }
```

djoksimo (changes requested):

Tests/ForageSDKTests/ForageServiceTests.swift
dleis612: Agreed, but technically speaking isn't this test deterministic still? The random value is always going to be between `-0.025` and `0.025`.

djoksimo: The random value is always going to be between -0.025 and 0.025.

It's more handy to be able to fix the seed so that there isn't randomness in the functions that use `jitterAmount`, like `waitNextAttempt`. The `expectation` has the right maximum bound, but if we changed the jitter amount range from 25 ms to 300ms, the tests for `waitNextAttempt` could become flaky.

djoksimo: we could do something like this to ensure that `waitNextAttempt` has repeatable behavior:

```swift
internal func jitterAmountInSeconds(using rng: () -> Int = defaultRNG) -> Double {
    return Double(rng()) / 1000.0
}

private func defaultRNG() -> Int {
    return Int.random(in: -25...25)
}
```

Tests/ForageSDKTests/ForageServiceTests.swift
```diff
@@ -205,6 +205,41 @@ final class ForageServiceTests: XCTestCase {
...
+        let mockSession = URLSessionMock()
+        let service = LiveForageService(provider: Provider(mockSession))
+        
+        let failureExpectation = XCTestExpectation(description: "Completion called before delay")
+        let successExpectation = XCTestExpectation(description: "Completion called after delay")
+        
+        // Mark the expectation as fulfilled only if it failed
+        failureExpectation.isInverted = true
+        
+        service.waitNextAttempt {
```
dleis612, line 233: @djoksimo To remove rng from this function we would need to use mock injection to the `LiveForageService` class and keep a reference to the function passed into the `jitterAmountInSeconds` function. Kinda seems like overkill.

djoksimo:
* main
  * **PR #106** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/106" target="_blank">Graphite</a>  👈
    * **PR #116** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/116" target="_blank">Graphite</a> 
      * **PR #119** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/119" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/106?utm_source=stack-comment).

djoksimo (approved):

Sources/ForageSDK/Foundation/Network/LDManager.swift
```diff
@@ -11,12 +11,79 @@ import LaunchDarkly
...
+            default:
+                return 1000
+            }
+        }
+    }
+}
+
+internal protocol LDManagerProtocol {
+    func getVaultType(ldClient: LDClientProtocol?, genRandomDouble: () -> Double, fromCache: Bool) -> VaultType
+    func getPollingIntervals(ldClient: LDClientProtocol?) -> [Int]
```
djoksimo, line 86: can we mark `ldClient` as required if it's meant to be set by the caller?

@dleis612

djoksimo (changes requested):

Sources/ForageSDK/Foundation/Network/LDManager.swift
dleis612: Uhhhh you're probably right!

djoksimo: discussed via slack! We can leave it as-is

djoksimo (approved):

PR #110, opened 2023-09-28
[FX-594] Respond with user_error if invalid PIN is submitted

dleis612:
## What
When a user tries to submit a PIN of an invalid length, we can respond immediately with a user_error.

## Test Plan
I tested this locally and will be adding a qa-test for the behavior!

### DIFF
--- a/Sources/ForageSDK/Foundation/Network/Utils/Constants.swift
+++ b/Sources/ForageSDK/Foundation/Network/Utils/Constants.swift
@@ -1,0 +1,18 @@
+//
+//  File.swift
+//  
+//
+//  Created by Danny Leiser on 9/28/23.
+//
+
+import Foundation
+
+struct CommonErrors {
+    static let INVALID_PIN_ERROR = ForageError(errors: [
+        ForageErrorObj(
+            httpStatusCode: 400,
+            code: "user_error",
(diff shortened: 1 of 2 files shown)


Sources/ForageSDK/Foundation/Network/Utils/Constants.swift
```diff
@@ -0,0 +1,18 @@
...
+//  File.swift
+//  
+//
+//  Created by Danny Leiser on 9/28/23.
+//
+
+import Foundation
+
+struct CommonErrors {
+    static let INVALID_PIN_ERROR = ForageError(errors: [
```
djoksimo, line 11:
```suggestion
    static let INCOMPLETE_PIN_ERROR = ForageError(errors: [
```

djoksimo (changes requested):

Sources/ForageSDK/Foundation/Network/Utils/Constants.swift
dleis612: Whoops: [9bacb87](https://github.com/teamforage/forage-ios-sdk/pull/110/commits/9bacb87a2b5b7a0d64ebafa280fc5a2e849ca93b)

PR #112, opened 2023-10-03
[FX-388] Remove pinType field

dleis612:
## What
This field has no use anymore and can be removed. I am going to get in touch with GoPuff to ensure that they don't use the field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,19 +179,8 @@ private let foragePinTextField: ForagePINTextField = {
     let tf = ForagePINTextField()
     tf.borderRadius = 10
     tf.backgroundColor = .systemGray6
-    tf.pinType = .balance
     return tf
 }()
-```
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -179,21 +179,10 @@ private let foragePinTextField: ForagePINTextField = {
...
 ```
 
-To identify the type of pin we are handling in the component, you can use the `pinType` property. We have support for these types:
-
-```swift
-public enum PinType: String {
-    case snap
-    case nonSnap
-    case balance
-}
```
djoksimo, line 194: If we wanted to gracefully deprecate this without having to release a new major version, we could use the `@available` feature

```swift
@available(*, deprecated, message: "PinType is deprecated.")
public enum PinType: String {
```

dleis612: I don't think this is used by anybody and may just rip the silent code removal

djoksimo (approved):

PR #115, opened 2023-10-10
[FX-622] Add temp workaround for clearText BT bug

djoksimo:
## What

- Temp workaround for https://linear.app/joinforage/issue/FX-622/bug-make-basis-theory-pin-cleartext-not-crash-in-ios
- Clean up how we display error details in sample app

## Why

https://linear.app/joinforage/issue/FX-622/bug-make-basis-theory-pin-cleartext-not-crash-in-ios

## Test Plan

- Added unit tests for clearText
- Call `clearText` in sample app

### ⚠️ Should be released with #120 , but can be released as-is (PATCH).

This change will ideally be [followed by a "full" fix once Basis Theory releases an official](https://linear.app/joinforage/issue/FX-622/bug-make-basis-theory-pin-cleartext-not-crash-in-ios#comment-1f9863a8) `clearText` method.

### DIFF
--- a/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
+++ b/Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
@@ -152,7 +152,9 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
     // MARK: - Public API
     
     func clearText() {
-        textField.text = ""
+        DispatchQueue.main.async {
+            self.textField.text = ""
+        }
     }
     
     var borderWidth: CGFloat {
--- a/Sample/SampleForageSDK/Sections/CapturePayment/CapturePaymentView.swift
+++ b/Sample/SampleForageSDK/Sections/CapturePayment/CapturePaymentView.swift
@@ -134,6 +134,7 @@ class CapturePaymentView: UIView {
         ForageSDK.shared.capturePayment(
             foragePinTextField: inputFieldReference,
             paymentReference: paymentReference) { result in
+                inputFieldReference.clearText()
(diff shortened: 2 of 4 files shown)


djoksimo:
* main
  * **PR #115** <a href="https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/115" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-ios-sdk/115?utm_source=stack-comment).

Sources/ForageSDK/Component/ForagePINTextField/Vault/BtPINTextField.swift
```diff
@@ -152,7 +152,9 @@ class BasisTheoryTextFieldWrapper: UIView, VaultWrapper {
     // MARK: - Public API
     
     func clearText() {
-        textField.text = ""
+        DispatchQueue.main.async {
```
dleis612, line 155: Did you test that this crashed before and now this solves the issue? Just making sure we don't introduce a new issue

djoksimo: yep, tested!

removing the `DispatchQueue.main.async` causes this to crash:

image

inserting it makes it not crash
Also, nesting `DispatchQueue.main.async` is also safe/works, which means if GoPuff wrapped the call to `clearText` in their own `DispatchQueue.main.async`, it would also work 😄 

```swift
        DispatchQueue.main.async {
            DispatchQueue.main.async {
                self.textField.text = ""
            }
        }
```

dleis612 (approved):

PR #125, opened 2023-11-13
[FX-480] Add Polling unit tests

djoksimo:
## What

- Move `Polling` to its own service; the `LiveForageService.swift` file was dealing with too many concerns and was hard to test
- Add unit tests for `Polling`
- Return `ForageError` for polling errors, where we were previously returning string errors to the client.

## Why

https://linear.app/joinforage/issue/FX-480/add-unit-tests-for-ios-polling

## Test Plan

- ✅  Added an exhaustive suite of unit tests for `Polling`

## How

Can be released as-is.

### DIFF
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -46,7 +46,17 @@ public class ForageSDK {
 
         VGSCollectLogger.shared.disableAllLoggers()
         let provider = Provider(logger: logger)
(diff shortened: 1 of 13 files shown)


dleis612 (approved):
Can we change the name to `PollingService` rather than `LivePollingService`? Tiago came up with that naming convention and I've always disliked it haha.

If the qa-tests are passing with these changes, then everything looks good to me!

PR #126, opened 2023-11-14
[FX-658] Potential fix for LDClient not being initialized

djoksimo:
## What

- Initialize `LDManager` and consequently the `LDClient` at the earliest point in the SDK lifecycle (inside `ForageSDK.setup`) – instead of inside `ForageSDK.init`

## Why

Race condition that I think might have have been the cause:

1. Client calls `ForageSDK.setup(merchantID: ...)` in their app
2. ⏩  Client calls `let tf = ForagePINTextField()` in their app
3. SDK called `getVaultType(...)` ([here](https://github.com/teamforage/forage-ios-sdk/blob/896cabe60801c2bc54afa8886ce5f56640e25c13/Sources/ForageSDK/Component/ForagePINTextField/ForagePINTextField.swift#L138))
4. LaunchDarkly was not initialized by ForageSDK on time

⚠️ because the client did not call `ForageSDK.shared.<something>` yet

https://linear.app/joinforage/issue/FX-658/fix-ldclientget-ios-initialization-issue

## Test Plan

- [x] Added unit tests to make sure LD is being initialized
- [x] Added block `_ = ForagePINTextField()` to increase likelihood of catching potential race conditions in the sample app
- [x] Made sure logging to the right environment still works

## How

Can be released as-is.

### DIFF
--- a/Sample/SampleForageSDK/AppDelegate.swift
+++ b/Sample/SampleForageSDK/AppDelegate.swift
@@ -12,13 +12,6 @@ import UIKit
 @main
 class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
-        ForageSDK.setup(
-            ForageSDK.Config(
-                merchantID: ClientSharedData.shared.merchantID,
-                sessionToken: ClientSharedData.shared.sessionToken
-            )
-        )
-
         // Override point for customization after application launch.
         return true
     }
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -10,8 +10,8 @@ public class ForageSDK {
     // MARK: Properties
 
     private static var config: Config?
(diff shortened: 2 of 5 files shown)


Sample/SampleForageSDK/AppDelegate.swift
```diff
@@ -12,13 +12,6 @@ import UIKit
 @main
 class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
-        ForageSDK.setup(
-            ForageSDK.Config(
-                merchantID: ClientSharedData.shared.merchantID,
-                sessionToken: ClientSharedData.shared.sessionToken
-            )
-        )
```
djoksimo, line 20: I don't think this block was necessary for our sample app^

djoksimo: I removed it because calling `ForageSDK.setup` with empty values for `sessionToken` and `merchantID` was causing the logger and launchdarkly to initialize in the wrong environment

dleis612 (approved):

dleis612 (approved):

PR #128, opened 2023-11-23
[FX-758] Expose the EBT card state value

dleis612:
## What
Add the inferred EBT card's state as a field that can be referenced on the PAN element.

## Why
GoPuff would like to have the option to do custom logic based on which state an EBT card is from.

## Test Plan
Adjusted the unit tests and exposed field in the sample app.


Sources/ForageSDK/Foundation/Protocol/ForagePANElement.swift
```diff
@@ -0,0 +1,12 @@
...
+//  ForagePANElement.swift
+//
+//
+//  Created by Danny Leiser on 11/23/23.
+//  Copyright © 2023-Present Forage Technology Corporation. All rights reserved.
+//
+
+/// Added information that can be inferred from the card value.
+public protocol CardInformation {
+    var binValue: String? { get }
```
dleis612, line 11: @djoksimo @devinmorgan would love input on what we should call this field. What would be a good name that could correspond to other information for PayFac for example? I don't want to make it specific to EBT cards and I also want to avoid using the word "state" if possible just because that feels confusing.

djoksimo: Thank you for opening this discussion.

> I don't want to make it specific to EBT card

I think the derived US state will be EBT-specific and it's safe to add a note like "This value is only set (non-nil) for EBT cards."

----

- I think it's safe to drop the `Value` and `Information` suffixes
- `Card` and `iin` align with conventions seen in other payments libraries.
- IIN is used more commonly throughout our client libraries and I believe we generally opt for "IIN" > "BIN"
- I've added a proposed suggestion for the `Card`'s comment
- I've added a comment for the `iin` property
- I think we should expose a public `USState` enum as seen below 👇🏻  (this may involve refactoring the `ForagePANValidations.swift` class

```suggestion
/// Represents a state of the United States of America.
public USState: String {
    case alabama = "AL"
    case alaska = "AK"
    case arizona = "AZ"
    case arkansas = "AR"
    ....
}

/// Represents card information based on the user's Primary Account Number (PAN) input.
public protocol Card {
    /// ------ IF WE WANT TO/CAN EXPOSE THE IIN to the client, other libraries do this for credit cards --------
    /// An acronym for Issuer Identification Number (IIN), also known as BIN (Bank Identification Number).
    /// The IIN is the first 6 digits of the PAN.
    iin: String? { get }
    
    /// The US state that issued the EBT card, derived from the Issuer Identification Number (IIN), 
    /// also known as BIN (Bank Identification Number).
    /// The IIN is the first 6 digits of the PAN.
    var usState: USState? { get }
```

Thoughts?

devinmorgan: Trying to adopt an MVP stance on this, what do you guys think about just exposing the `iin: String` and not exposing the `usState: USState` for now, at least? Reasons to consider this:

1. Telling GoPuff the card's IIN would be sufficient for them to gracefully navigate the Massachusetts issues since they could choose to handle cards with Massachusetts' IIN (`600875`) how they wish
2. We can always decide to add in `usState` later on if `iin` is found to be sufficiently useful.
3. My understanding is that IIN is relevant beyond just the scope of EBT cards. So, as we add more payment methods, IIN may continue to be relevant whereas (again just my understanding) `usState` may be EBT / benefits specific (e.g. `usState` doesn't matter if Forage ever processes credit cards in house)

What's your guys' take?

Sources/ForageSDK/Component/ForagePANTextField/ForagePANTextField.swift
```diff
@@ -10,7 +10,7 @@ public enum CardType: String {
     case EBT = "ebt"
 }
 
-public class ForagePANTextField: UIView, Identifiable, ForageElement, ForageElementDelegate {
+public class ForagePANTextField: UIView, Identifiable, ForageElement, CardInformation, ForageElementDelegate {
```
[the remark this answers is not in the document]
devinmorgan: > thoughts on introducing e a `derivedCardInfo` member rather than conforming `ForagePANTextField`to`CardInformation`?

This would be my vote as well because I think it organizes these fields in a logical and extensible way 👍 

As far as naming goes:
* `cardMetaInfo`
* `cardMeta`

I'm not too opinionated here

Sources/ForageSDK/Foundation/Protocol/ForagePANElement.swift
```diff
@@ -0,0 +1,74 @@
...
+//  Copyright © 2023-Present Forage Technology Corporation. All rights reserved.
+//
+
+/// Represents card information derived from the user's current Primary Account Number (PAN) input value.
+public protocol CardInformation {
+    
+    /// The US state that issued the EBT card, derived from the Issuer Identification Number (IIN),
+    /// also known as BIN (Bank Identification Number).
+    /// The IIN is the first 6 digits of the PAN.
+    var usState: USState? { get }
```
devinmorgan, line 15: I touched on this in one of the existing threads as well but I wonder if a simpler initial modification could be only exposing `iin` instead of `usState` since `iin` could apply beyond just EBT cards and `iin` technically communicates the US State if you have access to the list of US State IINs. We could always add a `usState` field later if we want.

devinmorgan (approved):
My understanding of this PR is that we'll expose the `usState` on the PAN Field and rather than exposing a string, we are trying to reduce errors / friction by exposing a public enum of the US states / territories. If that understanding is correct, this PR looks good to me!

djoksimo (changes requested):

Sources/ForageSDK/Foundation/Protocol/ForagePANElement.swift
```diff
@@ -0,0 +1,77 @@
...
+    case florida = "FL"
+    case georgia = "GA"
+    case guam = "GU"
+    case hawaii = "HI"
+    case idaho = "ID"
+    case illinois = "IL"
+    case indiana = "IN"
+    case iowa = "IA"
+    case kansas = "KS"
+    case kentucky = "KT"
```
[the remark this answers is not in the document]
dleis612: Lol good catch

djoksimo (approved):
Thank you @dleis612 !

PR #129, opened 2023-11-27
[FX-759] Call clearText on the main thread

djoksimo:
## What

- ensure `pin.clearText()` is called on main thread for all vault providers
- call `checkBalance` and `capturePayment` on concurrent background threads

## Test Plan

## How

Can be released as-is

### DIFF
--- a/Sample/SampleForageSDK/Sections/CapturePayment/CapturePaymentView.swift
+++ b/Sample/SampleForageSDK/Sections/CapturePayment/CapturePaymentView.swift
@@ -130,12 +130,14 @@ class CapturePaymentView: UIView {
 
         let inputFieldReference = isEbtSnap ? snapTextField : nonSnapTextField
 
-        ForageSDK.shared.capturePayment(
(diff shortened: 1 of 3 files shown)


dleis612 (commented):
I think you either need to update the single call in `ForagePINTextField` or update both calls in `VgsPINTextField` and `BtPINTextField`

djoksimo:
@dleis612 

> I think you either need to update the single call in `ForagePINTextField` or update both calls in `VgsPINTextField` and `BtPINTextField`

It's harder to unit test `ForagePINTextField` because it depends on calling `LDManager.shared.getVaultType(...)`
In a future PR, I will clean this up to make it easier to test `ForagePINTextField`

We've seen issues surrounding `clearText` in the past; it's safer to add _extra guarantees_ that `clearText` runs on the main thread – even if it means nesting calls to `DispatchQueue.main.async` – which is safe to do

---

Once i make it easier to stub `getVaultType`, I can refactor to a single call to `DispatchQueue.main.async`

PR #153, opened 2024-01-26
Pin the VGS Collect package to avoid build failures in CI

dleis612:
## What
Because of the optimistic operator we are seeing build failures on our automated CI tests run on main. VGS released version 1.16.0 yesterday which seems to be causing the issue. Our tests passed on the version before, which is 1.15.3.


djoksimo (approved):
NICE CATCH!

PR #156, opened 2024-02-04
Adjust vault endpoints to be synchronous

dleis612:
## What
Use new synchronous vault endpoints.


Sources/ForageSDK/Foundation/Network/Model/PaymentModel.swift
```diff
@@ -80,6 +80,28 @@ public struct PaymentModel: Codable {
     public let merchantFixedSettlement: String?
     public let platformFixedSettlement: String?
     public let refunds: [String]
+    
+    init(from model: RawPaymentResponseModel) {
```
dleis612, line 84: Test out adding errors here as an internal field

dleis612: This change works https://github.com/teamforage/forage-ios-sdk/pull/160/commits/4019a2b7649cd24a4635971bf671ffb0d942ef20

djoksimo (approved):

PR #157, opened 2024-03-29
[FX-1183] Pin all endpoints that don't require a version to default

dleis612:
## What
Add a default version header to all endpoints in the SDK to avoid backwards compatibility behavior that can be introduced by setting an API version header in the dashboard.

## Test Plan
Let QA tests run.


djoksimo (approved):

dleis612:
@djoksimo I was able to validate that a Tenant with a default API version header set experienced the generic endpoint behavior for all actions.

I also confirmed that the endpoints with defined API version headers set retained their override (for example, `/payment_methods` still has `API-VERSION: 2023-05-15`.

djoksimo (approved):

PR #158, opened 2024-03-29
[FX-1230] Add privacy manifest file to iOS SDK

evanfreeze:
## What
Adds Privacy Manifest file to iOS SDK. More details from Apple [here](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files).

Also bumps LaunchDarkly to version 9.6.0, the latest as of this PR. We technically only need to go to [9.4.0 to get their privacy manifest](https://github.com/launchdarkly/ios-client-sdk/pull/334), but looking through the changelog it doesn't seem like a huge bump to go from to 9.6.0 from there, so might as well just jump to the most recent version.

## Why
Since we use [required reasons APIs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api) in our SDK, we have to include this before May 1. I audited the use of each relevant API and captured the results in [this spreadsheet](https://docs.google.com/spreadsheets/d/1EssDinDQHC094kKeSqYCQ1cPw-wcL3dtUQzvwO8Mtjk/edit#gid=0).

## Test Plan

- ✅  Not a front-end change

## How
Can be merged as-is, no code changes.


Sources/ForageSDK/Foundation/Network/LDManager.swift
```diff
@@ -241,7 +241,7 @@ class LDManager: LDManagerProtocol {
     }
 
     private func createLDConfig(for environment: Environment) -> LDConfig {
-        LDConfig(mobileKey: getLDMobileKey(environment).rawValue)
+        LDConfig(mobileKey: getLDMobileKey(environment).rawValue, autoEnvAttributes: .enabled)
```
evanfreeze, line 244: This is the only code change. LaunchDarkly v9.x requires this second argument to `LDConfig` to get additional environment attributes. I went with their example (passing `.enabled`), but we can also do `.disabled` if we don't want this. More details in the [8.x->9.0 migration guide](https://docs.launchdarkly.com/sdk/client-side/ios/migration-8-to-9?_gl=1%2A1pcqasx%2A_gcl_au%2ANTY3Mjg4ODU3LjE3MTE3NDc5Nzk.).

ForageSDK.podspec
```diff
@@ -1,18 +1,18 @@
 Pod::Spec.new do |spec|
 
   spec.name         = "ForageSDK"
-  spec.version      = "4.4.2"
```
dleis612, line 4: There are 2 other locations where we update the version number:
- The README.md
- ForageSDK.swift

Sources/ForageSDK/Foundation/Network/LDManager.swift
dleis612: Seems harmless enough and something we may find useful in the future. I say we keep it.

ForageSDK.podspec
evanfreeze: Just pushed commits updating it in those places as well 👍🏻

dleis612 (approved):

PR #161, opened 2024-05-07
[FX-1351] Relax optional PaymentModel fields for deferred capture

djoksimo:
## What

- Make the following optional/nullable:

```swift
Payment.amount
Payment.delivery_address
Payment.is_delivery
```

## Why

https://linear.app/joinforage/issue/FX-1351/ios-relax-paymentmodel-for-deferred-flow

## Test Plan

- ✅ Tested manually against "empty"/"thin" and "fully-loaded" payment.

## How

Should be released asap, will be followed by improved testing flow for empty payments.


dleis612 (approved):

djoksimo:
Closed because we'll follow a different approach instead of relaxing the "usually-non-null" fields like `Payment.amount`, ....

PR #170, opened 2024-06-12
[1493] Only log errors when the response code is in the 500s

dleis612:
## What
The current iOS logs are pretty good. We store a ton of relevant data in the `customer_perceived_response` log that can be used to build monitoring. I would argue that we may not need to "error" when the http response code is in the 400s.
Screenshot 2024-06-12 at 11 56 06 AM
Screenshot 2024-06-12 at 11 56 35 AM
Screenshot 2024-06-13 at 11 13 23 AM

### DIFF
--- a/Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
+++ b/Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
@@ -283,7 +283,8 @@ extension ForageSDK: ForageSDKService {
     /// Determine if we should log an error or a warning
     private func logErrorResponse(_ message: String, error: Error?, attributes: [String: Encodable]?) {
         let statusCode = (error as? ForageError)?.httpStatusCode ?? 500
-        if (statusCode >= 400 && statusCode < 500) {
+        let isWarningLevelStatusCode = [400, 429].contains(statusCode)
+        if (isWarningLevelStatusCode) {
             ForageSDK.logger?.warn(message, error: error, attributes: nil)
         } else {
             ForageSDK.logger?.error(message, error: error, attributes: nil)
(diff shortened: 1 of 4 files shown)


Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
```diff
@@ -279,4 +279,14 @@ extension ForageSDK: ForageSDKService {
         )
         return false
     }
+    
+    /// Determine if we should log an error or a warning
+    private func logErrorResponse(_ message: String, error: Error?, attributes: [String: Encodable]?) {
+        let statusCode = (error as? ForageError)?.httpStatusCode ?? 500
+        if (statusCode >= 400 && statusCode < 500) {
```
dleis612, line 286: @djoksimo How do you feel about this? It's just a semantic change at the end of the day.

djoksimo: I worry that this would let through 422, 403 and other scary status codes.
I believe `expired_session_token` is a status 401, but it's your call if you want to put that in the "warning" list

```suggestion
        let isWarningLevelStatusCode = [400, 429].contains(statusCode)
        if (isWarningLevelStatusCode) {
```

dleis612: I thought about this as well and my original thinking was that none of the 400s are tooooooo scary... but I'm changing course. Let's adopt your thinking in the beginning and change if we feel it's too noisy.

dleis612: [528a3e5](https://github.com/teamforage/forage-ios-sdk/pull/170/commits/528a3e57f1b42d1a0fc7e3c4352cc3ebd37ad8f5)

djoksimo (approved):
Thank you sir!

PR #171, opened 2024-06-15
[FX-1509] Fix isFirstResponder state update in rosetta PIN input

evanfreeze:
## What
Fixes an issue where `isFirstResponder` on `RosettaPINTextField` wasn't getting updated properly when the underlying UITextField was focused or unfocused. Also adds a test for this.

Also fixes an issue with the Response Monitor unit tests that was introduced in https://github.com/teamforage/forage-ios-sdk/pull/170

## Why
https://linear.app/joinforage/issue/FX-1509/fix-isfirstresponder-state-update

## Test Plan
Manually verified the state updates in the sample app, also added a unit test.

## How
Can be merged as-is

### DIFF
--- a/Tests/ForageSDKTests/ForagePINTextFieldTests.swift
+++ b/Tests/ForageSDKTests/ForagePINTextFieldTests.swift
@@ -434,19 +434,39 @@ final class ForagePINTextFieldTests: XCTestCase {
         let rosettaPINTextField = RosettaPINTextField()
         let mockDelegate = MockDelegate()
         let textField = UITextField()
-
+        let window = UIWindow()
+        
         rosettaPINTextField.delegate = mockDelegate
+        window.addSubview(rosettaPINTextField)
         
         // textFieldDidChange
         rosettaPINTextField.textFieldDidChange(textField)
         XCTAssertEqual(mockDelegate.textFieldDidChangeCalledTimes, 1)
         
         // firstResponderDidChange
+        XCTAssertFalse(rosettaPINTextField.isFirstResponder)
+
+        // simulating the system firing the .editingDidBegin event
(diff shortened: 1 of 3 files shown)


Tests/ForageSDKTests/ForagePINTextFieldTests.swift
```diff
@@ -415,6 +415,39 @@ final class ForagePINTextFieldTests: XCTestCase {
...
+        // textFieldDidChange
+        rosettaPINTextField.textFieldDidChange(textField)
+        XCTAssertEqual(mockDelegate.textFieldDidChangeCalledTimes, 1)
+        
+        // firstResponderDidChange
+        rosettaPINTextField.editingBegan(textField)
+        XCTAssertEqual(mockDelegate.firstResponderDidChangeCalledTimes, 1)
+        
+        rosettaPINTextField.editingEnded(textField)
+        XCTAssertEqual(mockDelegate.firstResponderDidChangeCalledTimes, 2)
```
djoksimo, line 449: can we also assert the state of `pinTextField.isFirstResponder, true | false` in this test suite?

djoksimo (approved):

Tests/ForageSDKTests/ForagePINTextFieldTests.swift
evanfreeze: added in [2b230cf](https://github.com/teamforage/forage-ios-sdk/pull/171/commits/2b230cf1351cdcfb65e251320f8123b05d4ccda5) ✅

djoksimo (approved):
Thank you for fixing this and for including some additional tests + safeguards! @evanfreeze

PR #177, opened 2024-07-17
Allow .setup to update the merchantID and session token

djoksimo:
## What

- Make some methods `static` instead of `class` functions
- Remove dangerous references to `ForageSDK.shared.` that cause the logger to use the wrong environment
- Update sample app to cover the `.setup()` updates.

## Test Plan

- [x] Added unit test for `ForageSDK.setup`
- [x] Try calling `.setup` and different stages of the render lifecycle
- [x] Make sure logs appear where they're supposed to appear
- [ ] Make sure feature flags load as expected

## How

Can be released as-is. A clean up + refactoring ticket will be created.

### DIFF
--- a/Sample/SampleForageSDK/ViewController.swift
+++ b/Sample/SampleForageSDK/ViewController.swift
@@ -30,11 +30,14 @@ class ViewController: UIViewController {
 
         ClientSharedData.shared.merchantID = merchantID
         ClientSharedData.shared.sessionToken = sessionToken
-
+        
+        /// CHANGE WITH CAUTION
+        /// we intentionall call ``ForageSDK.setup``` with the wrong values first to ensure that we can update
+        /// the config values later on in the render (
         ForageSDK.setup(
             ForageSDK.Config(
-                merchantID: merchantID,
-                sessionToken: sessionToken
+                merchantID: "invalid mid/...",
+                sessionToken: "invalid session Token with no environment prefix"
             )
         )
     }
--- a/Sample/SampleForageSDK/ViewController.swift
+++ b/Sample/SampleForageSDK/ViewController.swift
@@ -32,11 +32,12 @@ class ViewController: UIViewController {
         ClientSharedData.shared.sessionToken = sessionToken
         
         /// CHANGE WITH CAUTION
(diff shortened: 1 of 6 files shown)


Sample/SampleForageSDK/ViewController.swift
```diff
@@ -30,11 +30,15 @@ class ViewController: UIViewController {
...
+        /// the config values later on in the ``CardNumberViewController.viewDidAppear`` method
         ForageSDK.setup(
             ForageSDK.Config(
-                merchantID: merchantID,
-                sessionToken: sessionToken
+                merchantID: "invalid mid/...",
+                // intentionally omitted environment prefix from this sessionToken
+                sessionToken: "invalid session Token with no environment prefix"
             )
         )
```
djoksimo, line 43: We can also add a test case in the Sample app that simply goes back and re-submits new session token + merchant ID, but that will take some time to introduce

So in the meantime, let's keep this code - unless @dleis612 objects

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -19,6 +19,9 @@ public class ForageSDK {
     public var environment: Environment = .sandbox
     // Don't update! Only updated when releasing.
     public static let version = "4.4.7"
+    
+    /// Until we migrate away from the singleton pattern in `ForageSDK`
+    /// ...Reference `shared` with caution! It can have dangerous side effects
```
devinmorgan, line 24: Ooooo didn't realize we were considering this - very exciting!

Tests/ForageSDKTests/ForageSDKTests.swift
```diff
@@ -34,6 +34,19 @@ final class ForageSDKTests: XCTestCase {
 
         XCTAssertEqual(MockForageSDK.shared.sessionToken, "dev_authToken123")
     }
+    
+    func testSetup_shouldUpdateSessionTokenAndMerchantID() {
```
dleis612, line 38: TODO: Add a test of setting the config, reading `ForageSDK.shared.environment`, then setting the config again.

djoksimo:
Closing this PoC in favor of Danny's upcoming changes

PR #178, opened 2024-07-18
[FX-1598] Make setup update merchant ID + session token

djoksimo:
## What

- [x] Make `ForageSDK.setup(...)` update merchant ID + session token

## Why

- Conventionally, merchants use `updateMerchantID` and `updateSessionToken`

## Test Plan

- [x] Manually verified that environment-specific things like logs and launchdarkly work as expected
- [x] Updated unit tests
- [x] Manually verified that merchant IDs and session tokens get updated as expected!
- [ ] Will follow-up with E2E mobile-qa-tests

## How

Should be released ASAP! The `ForageSDK.swift` class could use a lot of TLC 😅 , but this is a stopgap solution

### DIFF
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -12,8 +12,8 @@ public class ForageSDK {
     private static var config: Config?
     static var logger: ForageLogger?
     var service: ForageService?
-    var merchantID: String = ""
-    var sessionToken: String = ""
+    var merchantID: String = config?.merchantID ?? ""
+    var sessionToken: String = config?.sessionToken ?? ""
     var traceId: String = ""
 
     public var environment: Environment = .sandbox
@@ -28,10 +28,6 @@ public class ForageSDK {
             assertionFailure("ForageSDK is not initialized - call ForageSDK.setup() before accessing ForageSDK.shared")
             return
         }
-        environment = Environment(sessionToken: config.sessionToken)
-        merchantID = config.merchantID
-        sessionToken = config.sessionToken
-
         traceId = ForageSDK.logger?.getTraceID() ?? ""
 
         VGSCollectLogger.shared.disableAllLoggers()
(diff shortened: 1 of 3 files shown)


Sources/ForageSDK/ForageSDK.swift
```diff
@@ -28,10 +28,6 @@ public class ForageSDK {
             assertionFailure("ForageSDK is not initialized - call ForageSDK.setup() before accessing ForageSDK.shared")
             return
         }
-        environment = Environment(sessionToken: config.sessionToken)
```
djoksimo, line 31: `updateSessionToken` inside of `setup` updates the environment. The new unit tests will help catch any regressions/issues around environment-setting!

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -12,8 +12,8 @@ public class ForageSDK {
     private static var config: Config?
     static var logger: ForageLogger?
     var service: ForageService?
-    var merchantID: String = ""
-    var sessionToken: String = ""
+    var merchantID: String = config?.merchantID ?? ""
+    var sessionToken: String = config?.sessionToken ?? ""
```
djoksimo, line 16: ty @dleis612 for proposing this!

dleis612: There is the downside of this approach that these values are randomly the empty string, which may be hard for us to debug in the future. I think we should just put this down as tech debt to be solved later.

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -28,10 +28,6 @@ public class ForageSDK {
             assertionFailure("ForageSDK is not initialized - call ForageSDK.setup() before accessing ForageSDK.shared")
```
dleis612, line 28: Change this `guard` to just check to ensure the config exists.

dleis612 (changes requested):

Sources/ForageSDK/ForageSDK.swift
djoksimo: 5a61927098e171b0e90ad532184623d74f6c23da

devinmorgan (approved):
I know what it means to be in stopgap mode- approved!

dleis612 (approved):

PR #185, opened 2024-08-15
Refactor instance properties in ForageSDK to avoid manually syncing values

evanfreeze:
## What
Refactors some of the instance properties (merchantId, sessionToken, etc.) in `ForageSDK` to be derived from their static counterparts rather than be mutable, manually-set values. 

This guarantees the instance values in the `shared` singleton are always synced to what's in the static `config` property.

## Why
Reduces the likelihood of us forgetting to update something and creating unexpected issues.

## Test Plan

- ✅ Not a front-end change
- ✅ Unit Tests passed locally

## How
Can be safely merged into the `evan/sdk-hardening` branch, where I'm building up some refactoring changes.

### DIFF
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -63,6 +63,11 @@ public class ForageSDK {
     public struct Config {
         var merchantID: String
         var sessionToken: String
+        
+        public init(merchantID: String, sessionToken: String) {
+            self.merchantID = merchantID
+            self.sessionToken = sessionToken
+        }
     }
 
     /// Configures the Forage SDK with the given configuration.
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -63,7 +63,7 @@ public class ForageSDK {
     public struct Config {
         var merchantID: String
         var sessionToken: String
-        
+
         public init(merchantID: String, sessionToken: String) {
             self.merchantID = merchantID
             self.sessionToken = sessionToken
--- a/Sources/ForageSDK/ForageSDK.swift
+++ b/Sources/ForageSDK/ForageSDK.swift
@@ -6,28 +6,42 @@
 import Foundation
 
 public class ForageSDK {
-    // MARK: Properties
+    // MARK: - Static Properties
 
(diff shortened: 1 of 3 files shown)


Sources/ForageSDK/ForageSDK.swift
```diff
@@ -6,28 +6,42 @@
...
+    }
+    var merchantID: String {
+        ForageSDK.config?.merchantID ?? ""
+    }
+    var sessionToken: String {
+        ForageSDK.config?.sessionToken ?? ""
+    }
+    var traceId: String {
+        ForageSDK.logger?.getTraceID() ?? ""
+    }
```
evanfreeze, line 35: These computed properties work by running the closure code whenever they're accessed. So reading `ForageSDK.shared.merchantId` will run `ForageSDK.config?.merchantId ?? ""` under the hood, ensuring it's always using the latest value. 

It also makes these properties readonly since they aren't actually storing a value themselves — 
CleanShot 2024-08-15 at 12 13 22@2x

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -36,6 +50,8 @@ public class ForageSDK {
             ldManager: LDManager.shared
         )
     }
+    
+    // MARK: - Public API
```
evanfreeze, line 54: Adding these marks gives us this helpful UI menu accessible from the top of Xcode 

https://github.com/user-attachments/assets/b335a4db-ffa4-4765-8b4f-fa7d89e3bc8a

Sources/ForageSDK/ForageSDK.swift
```diff
@@ -47,11 +63,6 @@ public class ForageSDK {
     public struct Config {
         var merchantID: String
         var sessionToken: String
-
-        public init(merchantID: String, sessionToken: String) {
-            self.merchantID = merchantID
-            self.sessionToken = sessionToken
-        }
```
evanfreeze, line 54: Swift structs get "member-wise initializers" (an `init` method that has each property as a parameter) by default, so we don't need to explicitly define this. 
CleanShot 2024-08-15 at 12 20 41@2x
CleanShot 2024-08-15 at 12 21 20@2x

dleis612 (approved):

djoksimo (approved):

PR #192, opened 2024-11-14
Fix Coverage GHA workflow

djoksimo:
- Update Pyhton version to v3.12.0 in GHA runner


devinmorgan:
Closing this since it seems we've  the Coverage workflow is working now

PR #195, opened 2024-12-10
Refactor to avoid completion handler being called more than once in failure paths

cltnschlosser:
In balance check code paths this is used with checked continuations which crash when completion is called more than once. This rewrites the code to be a bit more straightforward (in the SSL error case, the completion was called 3 times). I also enabled the `assertForOverFulfill` flag on the expectations, so the tests will catch regressions here.

### DIFF
--- a/Sources/ForageSDK/Foundation/Network/Provider.swift
+++ b/Sources/ForageSDK/Foundation/Network/Provider.swift
@@ -49,34 +49,29 @@ class Provider {
         do {
             let request = try endpoint.urlRequest()
             task = urlSession.dataTask(with: request) { data, response, error in
-                self.processResponse(response: response) { result in
-                    DispatchQueue.main.async {
-                        switch result {
-                        case .success:
-                            completion(.success(data))
-                        case .failure:
-                            if let data = data {
-                                self.processVaultData(
-                                    model: ForageServiceError.self,
-                                    code: nil,
-                                    data: data,
-                                    response: response
-                                ) { errorResult in
-                                    switch errorResult {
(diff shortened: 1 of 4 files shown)


Sources/ForageSDK/Foundation/Network/Provider.swift
```diff
@@ -91,31 +87,22 @@ class Provider {
...
-                    attributes: ["endpoint": httpResponse?.url?.path]
-                )
-                httpResponse = response
-            case let .failure(error):
-                self.logger?.error("Failed to process response", error: error, attributes: nil)
-                return completion(.failure(error))
-            }
+        if let error = error {
+            let httpResponse = response as? HTTPURLResponse
+            let wrappedError = NSError(domain: "Error: \(error)", code: httpResponse?.statusCode ?? 500, userInfo: nil)
```
devinmorgan, line 92: we removed the `?`s everywhere else. Can they be removed here too?

Sources/ForageSDK/Foundation/Network/Provider.swift
```diff
@@ -91,31 +87,22 @@ class Provider {
...
+            self.logger?.error("Failed to process error for \(self.getResponseUrlPath(httpResponse))", error: wrappedError, attributes: nil)
+            return completion(.failure(wrappedError))
```
devinmorgan, line 94: why `completion(.failure(wrappedError))` instead of the original `completion(.failure(error))`?

devinmorgan:
We incorporated these changes into #195. Thanks for doing this, Colton!!

PR #103, opened 2023-09-11
[FX-519] Auto-generate reference documentation

djoksimo:
## What

- Adds support for [Jazzy](https://github.com/realm/jazzy) reference documentation generator

## Why

- Linear ticket https://linear.app/joinforage/issue/FX-519/generate-and-host-ios-mvp-reference-docs-using-xcode-tools

### View docs here 👉🏻  https://clinquant-meringue-5947c5.netlify.app/

```bash
cd <path-to-ios-sdk-repo>

[sudo] gem install jazzy

jazzy # generate reference-docs/

open reference-docs/index.html # open the docs website locally

# can use services like Netlify Drop to quickly deploy to temporary website and share the new docs with others
```

- ✅ Not a front-end change
- Navigated through docs for all of the methods and members

## How

- Can be released as-is. 
- Will have a follow-up PR that automatically deploys the docs to github pages.

PR #200, opened 2025-03-12
FX-2065 add HSA / FSA payment sheet

jdgazlay:
## What

A fairly large change in loc size, but much of these changes are similar styles and needs for each field in the HSA form.

- add `ForagePaymentSheet` UIView to display and style fields
- add customizable fields to help adapt to client styles
- expose field errors & state through `ForagePaymentSheet`
- add `Maskable` and `Validatable` protocols for centralized logic and compositional setup

add fields to gather card details for HSA/FSA data
- card holder name
- card number
- expiration date
- cvv (security code)
- zip code

add familiar API pattern for consuming clients
- Set defaults to all fields in the `ForagePaymentSheet`
  - Example: `ForagePaymentSheet.borderWidth = 2.0`
- Override specific field's styles in the `ForagePaymentSheet`
  - Example: `ForagePaymentSheet.cardHolderNameTextField.borderWidth = 5.0`
- top level `ForagePaymentSheet` state
  - `isComplete`: true if all fields are complete and valid
  - `completionErrors`: dictionary containing any subfield with an `invalidError` 
    - e.g. one field named "cardZip" is incomplete - `["cardZip": PaymentSheetError.incomplete]`
  - `currentFirstResponder`: pointer to the current first responder
- `ForagePaymentSheet` sub field state
  - `isDirty`: data has been entered previously
  - `isTouched`: client has interacted with the field
  - `invalidError`: reason input is invalid for submission (PaymentSheetError)
  - `isComplete`: true if complete and ready for submission
  - `isValid`: true if field is valid for submission
  - `name`: custom set name for the client to reference the fields

Adds new PaymentSheetError enum
PaymentSheetError
- `invalidCardNumber`
  - returned when the card number does not pass the [Luhn algorithm](https://en.wikipedia.org/wiki/Luhn_algorithm)
- `invalidDate`
  - returned if expiration date is invalid (either not a valid date or expired
- `incomplete`
  - returned if the field is incomplete

initial state
image

during filling out
image

invalid card error state
image

invalid date error state
image

valid zipcode state
image

invalid full zipcode state
image

valid full zip code state
image

complete state
image

## Why

facilitate the use of HSA/FSA benefits

## Test Plan

- ❌ Not a front-end change
- ✅ Unit Tests passed locally


Sample/SampleForageSDK/Sections/HSA/HSAView.swift
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    private func updateState(state: PaymentSheetObservableState) {
+        for var field in forageHSAPaymentSheet.fields {
+            if !field.isValid && !field.isFirstResponder && field.isDirty {
+                field.borderColor = UIColor(red: 0.60, green: 0.26, blue: 0.19, alpha: 1.0)
+            } else {
+                field.borderColor = UIColor(red: 0.01, green: 0.26, blue: 0.19, alpha: 1.0)
+            }
+        }
```
jdgazlay, line 215: this is a good demonstration of how a client can decide how they want to highlight errored fields based on the field's state

Sources/ForageSDK/Component/ForagePaymentSheet/Fields/ForageCardNumber.swift
```diff
@@ -0,0 +1,299 @@
...
+    }
+}
+
+// MARK: - UIResponder methods
+
+extension ForageCardNumber {
+    /// Make `ForageCardNumber` focused.
+    @discardableResult override public func becomeFirstResponder() -> Bool {
+        enhancedTextField.becomeFirstResponder()
+    }
```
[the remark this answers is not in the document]
jdgazlay: this is already being returned with the swift implicit return

Sources/ForageSDK/Component/Common/CardCVV.swift
```diff
@@ -0,0 +1,83 @@
...
+        newText = String(newText.prefix(4))
+        
+        isComplete = validateText(newText)
+        
+        text = newText
+
+        if !newText.isEmpty {
+            addClearButton(isVisible: true)
+        } else {
+            addClearButton(isVisible: false)
```
nitinsampath, line 79: What's the clear button? Is this necessary?

Sources/ForageSDK/Component/Common/CardHolderName.swift
```diff
@@ -0,0 +1,78 @@
...
+//  Created by Jerimiah on 3/6/25.
+//  Copyright © 2025-Present Forage Technology Corporation. All rights reserved.
+//
+
+import UIKit
+
+class CardHolderName: FloatingTextField, ObservableState, Validatable {
+    var invalidError: (any Error)?
+    
+    var validators: [(String) throws -> (Bool)] = [{
```
nitinsampath, line 14: Do we need any other validators like checking for non alphabet and quote characters?

Sources/ForageSDK/Component/Common/CardNumber.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+
+import UIKit
+
+class CardNumber: FloatingTextField, ObservableState, Maskable, Validatable {
+    // MARK: - Properties
+    
+    /// Maskable Properties
+    var actualText: String = ""
+    private(set) var maskPattern: String = "#### #### #### ####"
```
nitinsampath, line 16: Does this mean the user needs to manually input spaces when entering their card or is that done automatically?

Sources/ForageSDK/Component/Common/CardNumber.swift
```diff
@@ -0,0 +1,115 @@
...
+    }
+    
+    private func textLengthValidator(_ text: String) throws -> Bool {
+        if text.count < 16 {
+            throw PaymentSheetError.inComplete
+        }
+        return true
+    }
+    
+    private func luhnsCheckValidator(_ cardNumber: String) throws -> Bool {
```
nitinsampath, line 62: just making sure that luhns check applies to HSA/FSA cards always?

Sources/ForageSDK/Component/Common/CardZipCode.swift
```diff
@@ -0,0 +1,84 @@
...
+//  CardZipCode.swift
+//  ForageSDK
+//
+//  Created by Jerimiah on 2/26/25.
+//  Copyright © 2025-Present Forage Technology Corporation. All rights reserved.
+//
+
+import UIKit
+
+class CardZipCode: FloatingTextField, ObservableState, Validatable {
```
nitinsampath, line 11: do we enforce that only numbers can be typed in here?

Sources/ForageSDK/Foundation/Protocol/Maskable.swift
```diff
@@ -0,0 +1,120 @@
...
+        setNewCursorPosition(cursorOffset)
+    }
+    
+    /// default function to determine if the next character in the UITextField is a maskPattern template field
+    func isNextCharacterMaskTemplate(offset cursorOffset: NSInteger) -> Bool {
+        cursorOffset > 0 && maskPattern[safe: cursorOffset - 1] != "#"
+    }
+
+    /// default function to set the cursor position to a new position
+    func setNewCursorPosition(_ newPositionOffset: NSInteger) {
```
nitinsampath, line 115: why would we want to set a new cursor position?

nitinsampath (commented):
My first pass at the review - mostly general comments to clarify

Sources/ForageSDK/Component/Common/CardZipCode.swift
```diff
@@ -0,0 +1,84 @@
...
+    @objc func textFieldDidChange() {
+        defer {
+            wasBackspacePressed = false
+            isEmpty = text?.isEmpty ?? true
+            forageDelegate?.textFieldDidChange(self)
+        }
+
+        guard var newText = text else { return }
+        
+        newText = String(newText.prefix(5))
```
dleis612, line 71: We will inevitably see users attempt full zip codes (ex. 88310-7241). It looks like we restrict it, but just calling it out if we want to specifically format for it

Sources/ForageSDK/Component/ForagePaymentSheet/ForagePaymentSheet.swift
```diff
@@ -0,0 +1,386 @@
...
+//  Created by Jerimiah on 2/27/25.
+//  Copyright © 2025-Present Forage Technology Corporation. All rights reserved.
+//
+
+import UIKit
+
+public enum PaymentSheetError: Error {
+    case invalidCardNumber
+    case invalidDate
+    case inComplete
```
dleis612, line 14: nit: can we make this `incomplete` rather than `inComplete`?

Sample/SampleForageSDK/Sections/HSA/HSAView.swift
```diff
@@ -0,0 +1,233 @@
...
+        button.accessibilityIdentifier = "bt_send_hsa_fsa_data"
+        button.isAccessibilityElement = true
+        return button
+    }()
+
+    // MARK: Fileprivate Methods
+
+    @objc fileprivate func sendInfo(_ gesture: UIGestureRecognizer) {
+//        tokenizeCardButton.showLoading()
+//        ForageSDK.shared.tokenizeEBTCard(
```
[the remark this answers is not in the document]
jdgazlay: yeah, this will likely be something different. I haven't implemented the network call and tokenization yet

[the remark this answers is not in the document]
jdgazlay: I'm going to delete this code and add it back in the next PR

Sample/SampleForageSDK/Sections/HSA/HSAView.swift
```diff
@@ -0,0 +1,233 @@
...
+            leading: contentView.safeAreaLayoutGuide.leadingAnchor,
+            bottom: contentView.safeAreaLayoutGuide.bottomAnchor,
+            trailing: contentView.safeAreaLayoutGuide.trailingAnchor,
+            centerXAnchor: contentView.centerXAnchor,
+            padding: .init(top: 0, left: 24, bottom: 8, right: 24),
+            size: .init(width: 0, height: 48)
+        )
+    }
+
+    private func updateButtonState(isEnabled: Bool, button: UIButton) {
```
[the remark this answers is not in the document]
jdgazlay: Ah it's not being used. This along with a few other functions in this HSAView snuck in through some copy/paste from the `CardNumberView`. I'm going to remove this stuff for now until the next PR

Sources/ForageSDK/Component/Common/CardCVV.swift
jdgazlay: This clear button adds a helper button to the input box to quickly clear the text. Not necessary, but is currently used in the `ForagePANTextField`

Sources/ForageSDK/Component/Common/CardHolderName.swift
jdgazlay: I was going to ask about this as well. I was wondering what characters are allowed in Stripe's API so we can conform to that. Going to take a look at that today

Sources/ForageSDK/Component/Common/CardNumber.swift
jdgazlay: exactly. each card typed into this element will be checked against this luhns check.

Sources/ForageSDK/Foundation/Protocol/Maskable.swift
jdgazlay: When entering something like a credit card the user is typing only numbers while we format it to the `maskPattern`. If they've entered "4242|" ("|" is the cursor position in this example) and then they type the next number "42424|" we need to add the space and move the cursor so the user is still inputing at the correct position. "4242 4|". If we don't move the cursor they'll end up with something like "4242 |4" which would be confusing.

Sources/ForageSDK/Component/Common/CardZipCode.swift
jdgazlay: Great point! I'll add a mask for this and let them add the full zip code if a user wants.

Sources/ForageSDK/Component/ForagePaymentSheet/ForagePaymentSheet.swift
jdgazlay: 🤦 Good nit IMO. I think I just got camelCase happy here

Sources/ForageSDK/Component/Common/CardZipCode.swift
jdgazlay: added support for the full postal code use

Sources/ForageSDK/Component/Common/CardHolderName.swift
jdgazlay: We talked about this in the weekly HSA sync and agree this should match whatever requirements our payfac (stripe) requirements are - currently none from what I can see in the API 
https://docs.stripe.com/api/payment_methods/create#create_payment_method-billing_details-name

Sources/ForageSDK/Component/Common/CardZipCode.swift
```diff
@@ -44,13 +43,24 @@ class CardZipCode: FloatingTextField, ObservableState, Validatable {
...
         addTarget(self, action: #selector(textFieldDidChange), for: .editingChanged)
+        validators = [textLengthValidator]
     }
     
-    // MARK: Methods
+    private func textLengthValidator(_ text: String) throws -> Bool {
+        if text.count < 5 {
+            throw PaymentSheetError.incomplete
+        } else if text.count > 5 && text.count < 9 {
+            throw PaymentSheetError.incomplete
```
dleis612, line 59: I think this is satisfied elsewhere in your code but gonna ask. Can someone enter more than 9 digits?

jdgazlay: It's restricted by the Mask, so text can't be longer than the defined mask. We could probably use a test for this though to ensure that stays true.

jdgazlay: I just added some additional tests for text input length
35c24815c5964ec624158726a951981edfcfa163

djoksimo:
PS: Given the changes to the public surfaces, I'd recommend generating a copy of the reference documentation locally (using [Jazzy](https://github.com/teamforage/forage-ios-sdk/pull/103)), and previewing what the public documentation, and which public interfaces, methods, ... would be exposed to clients.

This is particularly helpful for accidentally exposing internal methods, enums, vars, interfaces that may lead to name collisions between our SDK and the SDK client (e.g. Forage and the client both use a `ZipTextField` element can cause collisions).

This is also helpful to ensure that all of our public code is well-documented & commented, in accordance with [Apple's comment conventions](https://github.com/realm/jazzy?tab=readme-ov-file#supported-documentation-keywords)

Learn more about running Jazzy: https://github.com/teamforage/forage-ios-sdk/pull/103

CC @pixelitobenito for visibility, no action needed right now.

Sources/ForageSDK/Foundation/Protocol/Maskable.swift
```diff
@@ -0,0 +1,118 @@
...
+//  ForageSDK
+//
+//  Created by Jerimiah on 3/13/25.
+//  Copyright © 2025-Present Forage Technology Corporation. All rights reserved.
+//
+
+import UIKit
+
+// protocol to facilitate masking the text in a UITextField
+public protocol Maskable: UITextField {
```
djoksimo, line 12: does this need to be public?

Sources/ForageSDK/Component/Common/PoweredByForageImage.swift
```diff
@@ -0,0 +1,69 @@
...
+//  PoweredByForageImage.swift
+//  ForageSDK
+//
+//  Created by Jerimiah on 2/26/25.
+//  Copyright © 2025-Present Forage Technology Corporation. All rights reserved.
+//
+
+import UIKit
+
+class PoweredByForageImage: UIView {
```
djoksimo, line 11: @dleis612 @jdgazlay This has been historically too small to read on iOS. Flagging that we should eventually make the powered by Forage logo bigger for better brand recognition – as it would help build trust & reputation for the Forage brand.

jdgazlay: I thought it looked quite small myself, so I'm glad to hear I'm not off base there. I'll keep that in mind as we integrate HSA/FSA

Sources/ForageSDK/Foundation/Protocol/Maskable.swift
jdgazlay: Probably not. I'll remove from this and `Validatable`

pixelitobenito:
> CC @pixelitobenito for visibility, no action needed right now.

Seen! This'll be my first iOS SDK change, so when we know what exactly is changing docs-wise, let me know in #documentation. Earlier the better, as it might take me a bit to figure out the technical logistics of editing this content.

Sources/ForageSDK/Foundation/Protocol/Maskable.swift
jdgazlay: changed in https://github.com/teamforage/forage-ios-sdk/pull/200/commits/1dda41e2b8860018d23523439993d163eb121660

jdgazlay:
> PS: Given the changes to the public surfaces, I'd recommend generating a copy of the reference documentation locally (using [Jazzy](https://github.com/teamforage/forage-ios-sdk/pull/103)), and previewing what the public documentation, and which public interfaces, methods, ... would be exposed to clients.
> 
> This is particularly helpful for accidentally exposing internal methods, enums, vars, interfaces that may lead to name collisions between our SDK and the SDK client (e.g. Forage and the client both use a `ZipTextField` element can cause collisions). Prefixing public code with `Forage<InterfaceName>` is a great pattern to follow for this reason, which I see that you did ⭐
> 
> Auto-generating docs via Jazzy is also helpful to ensure that all of our public code is well-documented & commented, in accordance with [Apple's comment conventions](https://github.com/realm/jazzy?tab=readme-ov-file#supported-documentation-keywords)
> 
> Learn more about running Jazzy: #103
> 
> CC @pixelitobenito for visibility, no action needed right now.

I generated these docs and read through, and there's not anything exposed that I wouldn't expect based on the changes. I'll check again after we add the api calls in the next PR. 

@pixelitobenito I'll keep you in the loop in #documentation as things get hardened 👍

dleis612 (approved):

PR #225, opened 2026-07-07
Single PIN

frago12:
## What
- Add support for Single PIN feature on deferred payment capture
- Add form in the Sample app to manually test this

## Why
We want to add support for creating multiple deferred payment captures for different merchants in a method call, overriding the merchant id that was used to initialize the SDK.

### DIFF
--- a/Sample/SampleForageSDK/Sections/SinglePIN/SinglePINView.swift
+++ b/Sample/SampleForageSDK/Sections/SinglePIN/SinglePINView.swift
@@ -1,0 +1,431 @@
+//
+//  SinglePINView.swift
+//  SampleForageSDK
+//
+//  © 2022-2025 Forage Technology Corporation. All rights reserved.
+//
+
+import ForageSDK
+import Foundation
+import UIKit
+
+/// A single payment row: a merchant ref + SNAP amount form that creates one SNAP payment.
+/// After a successful creation it holds the resulting `DeferredPayment` so the parent screen
+/// can defer-capture it later with a single PIN.
+final class PaymentFormRow: UIView {
+    // MARK: Public Properties
+
+    /// The payment created from this row's inputs, or `nil` if it hasn't been created yet.
+    private(set) var deferredPayment: DeferredPayment?
+
+    /// Injected by the parent to perform the actual payment creation. Returns the created payment reference.
+    var onCreatePayment: ((_ merchantRef: String, _ amount: String, _ completion: @escaping (Result<String, Error>) -> Void) -> Void)?
+
+    // MARK: Private Components
+
+    private let accessibilityPrefix: String
+
+    private lazy var merchantRefTextField: UITextField = {
+        let tf = UITextField()
+        tf.placeholder = "Merchant ref"
+        tf.borderStyle = .roundedRect
+        tf.autocapitalizationType = .none
+        tf.autocorrectionType = .no
+        tf.accessibilityIdentifier = "\(accessibilityPrefix)_merchant_ref"
+        tf.isAccessibilityElement = true
+        return tf
+    }()
+
+    private lazy var amountTextField: UITextField = {
+        let tf = UITextField()
+        tf.placeholder = "SNAP amount"
+        tf.borderStyle = .roundedRect
+        tf.keyboardType = .decimalPad
+        tf.accessibilityIdentifier = "\(accessibilityPrefix)_amount"
+        tf.isAccessibilityElement = true
+        return tf
+    }()
+
+    private lazy var createButton: UIButton = .createPaymentButton(
+        title: "Create SNAP Payment",
+        accessibilityIdentifier: "\(accessibilityPrefix)_create",
+        fundingType: .ebtSnap,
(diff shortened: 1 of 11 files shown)


Sample/SampleForageSDK/Sections/SinglePIN/SinglePINView.swift
```diff
@@ -0,0 +1,431 @@
+//
+//  SinglePINView.swift
+//  SampleForageSDK
+//
+//  © 2022-2025 Forage Technology Corporation. All rights reserved.
```
jdgazlay, line 5:
```suggestion
//  © 2026 Forage Technology Corporation. All rights reserved.
```

Sample/SampleForageSDK/Sections/SinglePIN/SinglePINView.swift
```diff
@@ -0,0 +1,431 @@
...
+                self.resultLabel.text = ""
+                self.errorLabel.text = "Create both payments before deferring."
+                completion()
+            }
+            return
+        }
+
+        let payments = [firstPayment, secondPayment]
+
+        DispatchQueue.global(qos: .userInitiated).async {
```
jdgazlay, line 292: This dispatches the SDK call to a background queue, but `deferMultiPaymentCapture` synchronously reads `foragePinTextField.isComplete` (UIKit view state) via `validatePin`. So we're touching UI off the main thread. Let's drop the global dispatch and call it directly like the other screens. The SDK already hops to main internally where it reads the PIN.

Sources/ForageSDK/Foundation/ForageSDK+ForageServices.swift
```diff
@@ -309,6 +342,53 @@ extension ForageSDK: ForageSDKService {
...
+        }
+
+        let pinCollector = foragePinTextField.getPinCollector()
+
+        Task.init {
+            do {
+                // Defer each payment sequentially using the single collected PIN. One deferral is
+                // performed per element, so the same merchantID may repeat. Each payment's merchantID
+                // overrides the merchant ID set at SDK init.
+                for payment in payments {
```
jdgazlay, line 375: Fail-fast means a mid-batch failure leaves earlier payments already deferred server-side, but the completion only surfaces the raw first error. The caller can't tell which payment failed or which succeeded, so they can't retry just the failures. This is a fundamental API behavior we should fix before release.

A couple of options:

- include the failing `paymentReference` (and/or the successfully deferred refs) in a dedicated error type
- return a per-payment result instead of `Result<Void, Error>`

jdgazlay: The vault re-reads the PIN from the live text field on every submission, and each deferral is 3 sequential round trips, so for N payments the field has to stay untouched for 3N requests. If the user edits or clears the field mid-batch, later payments submit a different PIN or fail with `incomplete_pin`. At minimum let's document that callers shouldn't clear the field until completion. Disabling input for the duration would be even better.

jdgazlay (changes requested):
I have a few architectural concerns with this approach.

Architectural concerns:

- This makes 3 sequential requests per payment (GET payment -> GET payment method -> vault submit), so N payments = 3N round trips from what we know are often older devices in often low-network-coverage areas, with no retries, no timeout tuning, and no way to cancel.
- At ~97% per-request success on a weak connection, a 5-payment batch (15 requests) completes only about 2 out of 3 times.
- A mid-batch failure leaves earlier payments deferred with no way to tell the caller which ones succeeded or failed.
- Each iteration also redoes identical work server-side. The vault repeats the same DB + AWS cryptography operations N times to produce the same encrypted PIN block.

I suggest we do this through a batch endpoint:

- I traced this through the vault + payments-api and a batch endpoint (one PIN submission carrying the payment/merchant pairs) is genuinely cheap work, maybe roughly 1-2 weeks end-to-end work.
- We have an in-house precedent for it where the order-level `collect_pin` already does "one PIN, many payments" (it's single-merchant only, which is the one thing we'd need to generalize).
- This would collapse the client to 3 flat requests, let the server make the defer all-or-nothing, and eliminate the duplicated crypto.

If needed now for timing:

- The completion needs per-payment results (or at minimum the failing `paymentReference`). `Result<Void, Error>` is the part that becomes a breaking change later, so let's get the shape right before release.
- The PIN field's mutability needs to be addressed as well. The vault re-reads the field on every submission, so an edit mid-batch silently submits a different PIN for the remaining payments. Disabling interaction on the field for the duration of the call would likely cover it.

Sample/SampleForageSDK/Sections/SinglePIN/SinglePINViewController.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  SinglePINViewController.swift
+//  SampleForageSDK
+//
+//  © 2022-2025 Forage Technology Corporation. All rights reserved.
```
jdgazlay, line 5:
```suggestion
//  © 2026 Forage Technology Corporation. All rights reserved.
```

